PR #33, opened 2022-09-26
Feature/sbachmei/mic 3371 exposure to sbp medication

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3371](https://jira.ihme.washington.edu/browse/MIC-3371)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/concept_models/vivarium_us_cvd/concept_model.html#healthcare-system-modeling)

Implements the treatment ramp for sbp medication. Some notes:
* treatment effects are coming in the following PR
* ldlc medication implementation is coming in a following PR
* Some LDLC stuff is here b/c it was required for baseline medication coverage

### Verification and Testing
Ran through several time steps.


src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -241,9 +345,143 @@ def visit_doctor(
 
         return df
 
-    def update_treatment(self, index: pd.Index):
-        # TODO [MIC-3371, MIC-3375]
-        pass
+    def update_treatment(self, df: pd.DataFrame) -> pd.DataFrame:
```
albrja: This should return a dataframe and an index?

stevebachmeier: Good catch

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -241,9 +345,143 @@ def visit_doctor(
...
+            df[self.sbp_medication_column] + 1
+        )
+        df.loc[medication_change, "to_schedule"] = 1
+
+        return df
+
+    def get_measurement_error(self, index, mean, sd):
+        """Return measurement error assuming normal distribution"""
+        draw = self.randomness.get_draw(index)
+        return scipy.stats.norm(loc=mean, scale=sd).ppf(draw)
```
albrja: If you end up using this for multiple components this looks like a useful function you could throw into utilities.

albrja (approved):

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -171,25 +217,85 @@ def on_time_step_cleanup(self, event: Event) -> None:
...
+                    self.sbp_medication_column,
+                ]
+            ]
+        )
+
+    def calculate_medication_coverage_probabilities(self, df: pd.DataFrame) -> Dict:
+        """Determine the probability of each simulant being medicated"""
+        medication_coverage_covariates = {}
+        med_types = data_values.BASELINE_MEDICATION_COVERAGE_COEFFICIENTS._fields
+        for med_type in med_types:
```
rmudambi: You can iterate over a NamedTuple (it's a significant reason why we use them).

```
for c_int, c_sbp, c_ldlc, c_age, c_sex in data_values.BASELINE_MEDICATION_COVERAGE_COEFFICIENTS:
    ....
```

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -171,25 +217,85 @@ def on_time_step_cleanup(self, event: Event) -> None:
...
+            df[
+                [
+                    self.visit_type_column,
+                    self.scheduled_visit_date_column,
+                    self.sbp_medication_column,
+                ]
+            ]
+        )
+
+    def calculate_medication_coverage_probabilities(self, df: pd.DataFrame) -> Dict:
```
rmudambi: Why does this return a dict not a dataframe?

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -171,25 +217,85 @@ def on_time_step_cleanup(self, event: Event) -> None:
...
+            p=list(data_values.BASELINE_MEDICATION_LEVEL_PROBABILITY["sbp"].values()),
+        )
+        df.loc[medicated_ldlc, self.ldlc_medication_column] = self.randomness.choice(
+            medicated_ldlc,
+            choices=list(data_values.BASELINE_MEDICATION_LEVEL_PROBABILITY["ldlc"].keys()),
+            p=list(data_values.BASELINE_MEDICATION_LEVEL_PROBABILITY["ldlc"].values()),
         )
 
+        return df
+
```
rmudambi: You should be specific in your typing of `visitors`. I think it's `Dict[str, pd.Index]`?

That being said, I've commented above that I think it's better to set the visit type outside this function and just pass and index.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -79,20 +105,39 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
...
+        )
+        df[self.ldlc_medication_adherence_type_column] = self.randomness.choice(
+            index,
+            choices=list(data_values.MEDICATION_ADHERENCE_TYPE_PROBABILITIY["ldlc"].keys()),
+            p=list(data_values.MEDICATION_ADHERENCE_TYPE_PROBABILITIY["ldlc"].values()),
+        )
 
-        # Handle simulants initialized in an post/chronic state
-        visitors = {}
+        df = self.initialize_medication_coverage(df)
```
rmudambi: You have adopted a pattern here of generating a dataframe and then adding and removing columns from it. This makes it very hard to see what is actually changing in it when you go into and out of functions. You have created a very deep tree of function calls and it get very opaque as to what is happening in each one. Let's talk about this and see if there's a way to make this easier to read.

On a similar note, it's a good idea to have more descriptive variable names than this.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -207,31 +313,29 @@ def visit_doctor(
...
 
-        # Update treatments
+        # Update treatments (all visit types go through the same treatment ramp)
         to_visit = pd.Index([])
         attended_visit_types = [
             visit_type
-            for visit_type in data_values.VISIT_TYPES
-            if visit_type != data_values.VISIT_TYPES.MISSED
+            for visit_type in data_values.VISIT_TYPE
+            if not visit_type in [data_values.VISIT_TYPE.MISSED, data_values.VISIT_TYPE.NONE]
```
rmudambi: Why are you creating this data structure every time this function is called? Isn't it going to have the same values every time?

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -118,12 +162,10 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
...
             == models.ACUTE_MYOCARDIAL_INFARCTION_STATE_NAME
         )
-        emergency_not_already_scheduled = index[
-            (mask_acute_is | mask_acute_mi) & ~(mask_chronic_is | mask_post_mi)
-        ]
-        visitors[data_values.VISIT_TYPES.EMERGENCY] = emergency_not_already_scheduled
-        df.loc[emergency_not_already_scheduled] = self.visit_doctor(
-            df=df.loc[emergency_not_already_scheduled],
+        emergency = index[(mask_acute_is | mask_acute_mi)]
+        visitors[data_values.VISIT_TYPE.EMERGENCY] = emergency
```
rmudambi: Why don't you just set the visit type column here rather than doing this complicate passing around of dictionaries? Then you can just pass the index of the simulants who are attending the doctor to this function.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -241,9 +345,143 @@ def visit_doctor(
 
         return df
 
-    def update_treatment(self, index: pd.Index):
-        # TODO [MIC-3371, MIC-3375]
-        pass
+    def update_treatment(self, df: pd.DataFrame) -> Tuple[pd.DataFrame, pd.Index]:
```
rmudambi: This function appears to be called only in one place, so extracting it to a new function only serves to make the code harder to follow.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -241,9 +345,143 @@ def visit_doctor(
...
+        do_not_prescribe_mid_sbp = mid_sbp.difference(to_prescribe_mid_sbp)
+        # If no tx prescribed, still schedule a followup
+        df.loc[do_not_prescribe_mid_sbp, "to_schedule"] = 1
+        # If tx prescribed, it must be one_drug_half_dose; also schedule followup
+        df.loc[
+            to_prescribe_mid_sbp, self.sbp_medication_column
+        ] = data_values.MEDICATION_RAMP["sbp"][
+            data_values.SBP_MEDICATION_LEVEL.ONE_DRUG_HALF_DOSE
+        ]
+        df.loc[to_prescribe_mid_sbp, "to_schedule"] = 1
```
rmudambi: You can just say schedule all people with mid-sbp

src/vivarium_nih_us_cvd/components/healthcare_visits.py
rmudambi: This may be too late to be be worth mentioning, but it might have been simpler to have implemented this as if we had two separate decision trees - one for the treatement ramp and one for deciding whether or not to schedule a follow up.

`schedule_follow_up = already_on_treatment or sbp >=130`

I think this approach will allow you to eliminate a lot of the more confusing parts of the code.

rmudambi (changes requested):

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: My thinking was it was annoying having to pass in the index of people to have treatment updated as well as "remember" to switch all of their visit types outside of the function so I would just do it all with this dictionary of {visit_type: index}.

stevebachmeier: It also made it simpler/future-proofed to ensure that simulants who have more than a single visit type (specifically, "missed" their scheduled appointment but had a "background" visit) have both accounted for in the visit type column (needed for observation)

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: Ah, yeah that was leftover from when I commented out all decision points in the ramp tree

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: Yes, good call, I'll move it out

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: i like it - looking forward to talking

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: Oops, yeah I need to be better about actually running `pytype`. But yes, irrelevant if I end up moving the visit type functionality out

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: Wouldn't I lose control over what coefficients went with what covariates? The way I have it I loop through the fields (ie the covariate types: sbp, ldlc, both, and none) and
1. calculate the coefficients (I agree I could have done this more cleanly as you pointed out)
2. but then I also use the fields as keys to a dictionary to store the calculated covariates.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: Oof, just me being sloppy

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: Will confirm with Syl that she wants visitors with missed appointments to also be part of background random draws.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: We discussed keeping the HospitalVisits class strictly for assigning visit types and scheduling followups and then creating two new treatment ramps (one for sbp and one for ldlc) to deal with treatments.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: The reason for this was to deal with simulants who both miss their scheduled appt and also go for a background. We checked with Syl and she confirmed if a simulant misses a scheduled one it's fine to not allow them to go in for background. I will remove this dictionary nonsense.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
             "utilization_rate", background_utilization_rate, requires_columns=["age", "sex"]
         )
 
-        # Add columns
-        self.visit_type_column = data_values.VISIT_TYPE_COLUMN
-        self.scheduled_visit_date_column = data_values.SCHEDULED_VISIT_DATE_COLUMN
-        self.miss_scheduled_visit_probability_column = (
-            data_values.MISS_SCHEDULED_VISIT_PROBABILITY_COLUMN
-        )
+        # Columns
```
rmudambi: You already have these strings defined quite clearly as constants in the `data_values` file (I'm not sure that's the right place for constants like these, but we can have a broader discussion about that as a team), so I don't see a need to reassign them here. If you do want to have a local constant defined on `self` I prefer to define these constants in the init.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -27,6 +27,8 @@ def setup(self, builder: Builder) -> None:
         self.clock = builder.time.clock()
         self.step_size = builder.time.step_size()
         self.randomness = builder.randomness.get_stream(self.name)
+        self.sbp = builder.value.get_value("high_systolic_blood_pressure.exposure")
```
rmudambi: These pipeline names could be defined as constants in the same way you defined the column name constants

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
             requires_columns=columns_required,
+            requires_values=values_required,
         )
 
         # Register listeners
-        builder.event.register_listener("time_step__cleanup", self.on_time_step_cleanup)
+        builder.event.register_listener(
+            "time_step__cleanup",
+            self.on_time_step_cleanup,
+            priority=data_values.COMPONENT_PRIORITIES.HEALTHCARE_VISITS,
```
rmudambi: This constant name is fine since this component only registers a listener on time-step cleanup, but to be clear this is actually not a component priority. This is the priority for this component's time-step cleanup listener. If you had an on time-step listener there's no guarantee you'd want this priority.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
-        distributed. All simulants initialized in an acute state should be
-        scheduled a followup visit 3-6 months out, uniformly distributed.
-
-        For simplicity, do not assign background screenings on initialization.
+        """Initializes the visit type of simulants in an emergency state so
+        that they can be sent to the medication ramps.
 
-        A burn-in period allows for the observed simulation to start with more
-        realistic treatments and scheduled followups.
+        Note that scheduled visits are initialized in the Treatment componenet
```
rmudambi: This is no longer true?

albrja (approved):

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+                + c_sex * pop["sex"].map(data_values.BASELINE_MEDICATION_COVERAGE_SEX_MAPPING)
+            )
+        # Calculate probabilities of being medicated
+        p_medication = {}
+        p_denominator = sum(medication_coverage_covariates.values()) + 1
+        for med_type, cov in medication_coverage_covariates.items():
+            p_medication[med_type] = pd.Series(cov / p_denominator, name=med_type)
+        p_medication["none"] = pd.Series(1 / p_denominator, name="none")
+
+        return pd.concat(p_medication, axis=1)
```
albrja: Im not sure I understand completely the math that is happening here but the handling of objects looks good.

stevebachmeier: It's this kinda funky system of equations to calculate covariates and then the respective probabilities of folks being initialized on sbp medication, ldlc medication, both medications, or no medications: https://vivarium-research.readthedocs.io/en/latest/concept_models/vivarium_us_cvd/concept_model.html#initialization-parameters

But it should be fine, I verified this is working as expected mutliple times

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+            p=list(data_values.MEDICATION_ADHERENCE_TYPE_PROBABILITIY["sbp"].values()),
+        )
+        pop[self.ldlc_medication_adherence_type_column] = self.randomness.choice(
+            pop.index,
+            choices=list(data_values.MEDICATION_ADHERENCE_TYPE_PROBABILITIY["ldlc"].keys()),
+            p=list(data_values.MEDICATION_ADHERENCE_TYPE_PROBABILITIY["ldlc"].values()),
+        )
+
+        # Initialize medication coverage
+        pop[self.sbp_medication_column] = np.nan
```
rmudambi: Why are you initializing these columns as np.nan? These are not numerical columns. You should have a `NO_TREATMENT` ramp level that you use for people not on treatment.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
+        mask_acute_mi = (
+            pop[self.myocardial_infarction_state_column]
+            == models.ACUTE_MYOCARDIAL_INFARCTION_STATE_NAME
+        )
+        emergency = pop.index[(mask_acute_is | mask_acute_mi)]
+        pop.loc[emergency, self.visit_type_column] = data_values.VISIT_TYPE.EMERGENCY
+
+        # Schedule followups
+        # On medication and/or Post/chronic state 0-6 months out
+        on_medication = pop[
```
rmudambi: This check will need to change if you change the value designating untreated as i suggest below.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
+
+        # Update simulants initialized in an emergency state
+        mask_acute_is = (
+            pop[self.ischemic_stroke_state_column] == models.ACUTE_ISCHEMIC_STROKE_STATE_NAME
+        )
+        mask_acute_mi = (
+            pop[self.myocardial_infarction_state_column]
+            == models.ACUTE_MYOCARDIAL_INFARCTION_STATE_NAME
+        )
+        emergency = pop.index[(mask_acute_is | mask_acute_mi)]
```
rmudambi: Nit: parentheses are not needed

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
-        or followup visit. In the event that a simulant goes for an emergency or
-        background visit but already has a followup visit scheduled for the future,
-        keep that scheduled followup and do not schedule another one.
+        or followup visit and schedule followups. In the event that a simulant
+        already has a followup visit scheduled for the future, keep that scheduled
+        followup and do not schedule a new one or another one.
         """
-        df = self.population_view.get(event.index, query='alive == "alive"')
-        visitors = {}
+        pop = self.population_view.subview(
```
rmudambi: You only need to use `subview` during simulant initialization. Its purpose is to allow you to access the population view even if some of the columns specified to be in the population view haven't been created yet. Once you're in the main loop, all columns are guaranteed to exist, so you no longer need to use `subview`.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
         mask_acute_mi = (
-            df[models.MYOCARDIAL_INFARCTION_MODEL_NAME]
+            pop[self.myocardial_infarction_state_column]
             == models.ACUTE_MYOCARDIAL_INFARCTION_STATE_NAME
         )
         mask_emergency = mask_acute_is | mask_acute_mi
-        visit_emergency = df[mask_emergency].index
-        visitors[data_values.VISIT_TYPES.EMERGENCY] = visit_emergency
+        visit_emergency = pop[mask_emergency].index
+        pop.loc[visit_emergency, self.visit_type_column] = data_values.VISIT_TYPE.EMERGENCY
```
rmudambi: You can use a mask as the first argument of `.loc` btw. It doesn't have to be an index.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
 
-        # Scheduled visits
+        # Missed scheduled (non-emergency) visits (these do not get re-scheduled)
         mask_scheduled_non_emergency = (
-            (df[self.scheduled_visit_date_column] > (event.time - event.step_size))
-            & (df[self.scheduled_visit_date_column] <= event.time)
+            (pop[self.scheduled_visit_date_column] > (event.time - event.step_size))
```
rmudambi: Since you're setting the scheduled visit value to `pd.NaT` in line 190, you don't need the `> event.time - event.step_size` check

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
-        # Missed scheduled (non-emergency) visits (these do not get re-scheduled)
-        missed_visit = self.randomness.filter_for_probability(
-            scheduled_non_emergency,
-            df.loc[scheduled_non_emergency, self.miss_scheduled_visit_probability_column],
-        )  # pd.Index
-        visitors[data_values.VISIT_TYPES.MISSED] = missed_visit
-        df.loc[missed_visit, self.scheduled_visit_date_column] = pd.NaT  # no re-schedule
-        visit_scheduled = scheduled_non_emergency.difference(missed_visit)
-        visitors[data_values.VISIT_TYPES.SCHEDULED] = visit_scheduled
+        scheduled_non_emergency = pop[mask_scheduled_non_emergency].index
```
rmudambi: Is this the same as `pop.index[mask_scheduled_non_emergency]`? You should choose one method of converting a mask to an index and stick with it as it's confusing (and not pythonic) to do the same thing in different ways.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
+
+        # Scheduled visits
+        visit_scheduled = scheduled_non_emergency.difference(visit_missed)
+        pop.loc[visit_scheduled, self.visit_type_column] = data_values.VISIT_TYPE.SCHEDULED
 
-        # Background visits (for those who did not go for another reason (even missed screenings))
-        maybe_background = df.index.difference(visit_emergency.union(visit_scheduled))
+        # Background visits (for those who did not go for another reason or miss their scheduled visit)
+        maybe_background = pop.index.difference(
+            visit_emergency.union(visit_missed).union(visit_scheduled)
```
rmudambi: This can just be `visit_emergency.union(scheduled_non_emergency)`

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
-            df.loc[scheduled_non_emergency, self.miss_scheduled_visit_probability_column],
-        )  # pd.Index
-        visitors[data_values.VISIT_TYPES.MISSED] = missed_visit
-        df.loc[missed_visit, self.scheduled_visit_date_column] = pd.NaT  # no re-schedule
-        visit_scheduled = scheduled_non_emergency.difference(missed_visit)
-        visitors[data_values.VISIT_TYPES.SCHEDULED] = visit_scheduled
+        scheduled_non_emergency = pop[mask_scheduled_non_emergency].index
+        pop.loc[scheduled_non_emergency, self.scheduled_visit_date_column] = pd.NaT
+        visit_missed = scheduled_non_emergency[
+            self.randomness.get_draw(scheduled_non_emergency)
```
rmudambi: ***IMPORTANT***

Since you're using the same `RandomnessStream` throughout this component, you will get the exact same results each time you call it on a given time-step. You need to provide an `additional_key` parameter to ensure each time you call it you get unique results. That key should be a useful identifier of what this random draw is being used for.

This is an issue in many locations in this PR. You can Ctrl+F `self.randomness` to find them all.

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
-        # Only schedule a followup if a future one does not already exist
-        has_followup_scheduled = df.loc[
-            to_visit.intersection(
-                df[(df[self.scheduled_visit_date_column] > event_time)].index
-            )
+        # Schedule followups
+        visitors_high_sbp = all_visitors.intersection(
+            measured_sbp[measured_sbp >= data_values.SBP_THRESHOLD.LOW].index
+        )
+        visitors_on_sbp_medication = all_visitors.intersection(
```
rmudambi: Note: this check will need to be changed when you add an unmedicated value for this column

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -37,213 +39,213 @@ def setup(self, builder: Builder) -> None:
...
+            pop[pop[self.sbp_medication_column].notna()].index
+        )
+        visitors_not_on_sbp_medication = all_visitors.difference(visitors_on_sbp_medication)
+        # Schedule those on sbp medication or those not on sbp medication but have a high sbp
+        needs_followup = visitors_on_sbp_medication.union(
+            visitors_not_on_sbp_medication.intersection(visitors_high_sbp)
```
rmudambi: This can just be `visitors_on_sbp_medication.union(visitors_high_sbp)`

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+    def __repr__(self):
+        return "Treatment"
+
+    def setup(self, builder: Builder) -> None:
+        self.randomness = builder.randomness.get_stream(self.name)
+        self.sbp = builder.value.get_value("high_systolic_blood_pressure.exposure")
+        self.ldlc = builder.value.get_value("high_ldl_cholesterol.exposure")
+
+        # Columns
+        self.ischemic_stroke_state_column = models.ISCHEMIC_STROKE_MODEL_NAME
```
rmudambi: See comment above in `healthcare_visits.py`

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+        ]
+        columns_required = [
+            "age",
+            "sex",
+            self.ischemic_stroke_state_column,
+            self.myocardial_infarction_state_column,
+        ]
+
+        self.population_view = builder.population.get_view(
+            columns_required + columns_created + [self.visit_type_column]
```
rmudambi: What is special about the visit type column?

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+            self.on_time_step_cleanup,
+            priority=data_values.COMPONENT_PRIORITIES.TREATMENT,
+        )
+
+    def on_initialize_simulants(self, pop_data: SimulantData) -> None:
+        """Implements baseline medication coverage as well as adherence levels
+        Anyone initialized in an emergency (acute) state will enter the treatment
+        ramps.
+
+        This component also initializes scheduled followups because doing so
```
rmudambi: This is no longer true as the HealthCareVisits component does this

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+
+        # Send anyone in emergency state to medication ramp
+        mask_acute_is = (
+            pop[self.ischemic_stroke_state_column] == models.ACUTE_ISCHEMIC_STROKE_STATE_NAME
+        )
+        mask_acute_mi = (
+            pop[self.myocardial_infarction_state_column]
+            == models.ACUTE_MYOCARDIAL_INFARCTION_STATE_NAME
+        )
+        emergency = pop.index[(mask_acute_is | mask_acute_mi)]
```
rmudambi: No need to convert this to an index; a mask works perfectly well in `.loc`

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+                    self.ldlc_medication_column,
+                    self.ldlc_medication_adherence_type_column,
+                ]
+            ]
+        )
+
+    def initialize_medication_coverage(self, pop: pd.DataFrame) -> pd.DataFrame:
+        """Initializes medication coverage and adherence levels"""
+        # Initialize adherence levels
+        pop[self.sbp_medication_adherence_type_column] = self.randomness.choice(
```
rmudambi: This is fine for now, but we will want to revisit this when it comes time to do the polypill intervention and have adherence propensity columns here and adherence pipelines. 

Sorry, I should have caught this when we talked through the design

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+            != data_values.MEDICATION_ADHERENCE_TYPE.ADHERENT
+        ].index
+        pop.loc[medicated_sbp.intersection(sbp_non_adherent), self.sbp_medication_column] = 1
+        pop.loc[
+            medicated_ldlc.intersection(ldlc_non_adherent), self.ldlc_medication_column
+        ] = 1
+
+        return pop
+
+    def calculate_medication_coverage_probabilities(self, pop: pd.DataFrame) -> pd.DataFrame:
```
rmudambi: Rename to `calculate_initial_medication_coverage_probabilities`?

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+        """Determine the probability of each simulant being medicated"""
+
+        # Calculate the covariates
+        medication_coverage_covariates = {}
+        for (
```
rmudambi: This is cleaner and leverages your named tuples better:
```
for coefficients in data_values.MEDICATION_COVERAGE_COEFFICIENTS:
    medication_coverage_covariates[coefficients.NAME] = np.exp(
        coefficients.INTERCEPT
        + coefficients.SBP * self.sbp(pop.index)
        + coefficients.LDLC * self.ldlc(pop.index)
        + coefficients.AGE * pop["age"]
        + coefficients.SEX * pop["sex"].map(data_values.BASELINE_MEDICATION_COVERAGE_SEX_MAPPING)
    )
```

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+            medicated_ldlc.intersection(ldlc_non_adherent), self.ldlc_medication_column
+        ] = 1
+
+        return pop
+
+    def calculate_medication_coverage_probabilities(self, pop: pd.DataFrame) -> pd.DataFrame:
+        """Determine the probability of each simulant being medicated"""
+
+        # Calculate the covariates
+        medication_coverage_covariates = {}
```
rmudambi: You should be able to make this a DataFrame from the start rather than a dict? That would enable you to do the denominator division operation in a one line vectorized operation.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+        p_denominator = sum(medication_coverage_covariates.values()) + 1
+        for med_type, cov in medication_coverage_covariates.items():
+            p_medication[med_type] = pd.Series(cov / p_denominator, name=med_type)
+        p_medication["none"] = pd.Series(1 / p_denominator, name="none")
+
+        return pd.concat(p_medication, axis=1)
+
+    def on_time_step_cleanup(self, event: Event) -> None:
+        """Update treatments"""
+        pop = self.population_view.subview(
```
rmudambi: Don't need `.subview` here as we aren't in the initialization phase.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+            .intersection(adherent)
+            .intersection(not_already_max_medicated)
+        )
+        pop_medicated.loc[medication_change, self.sbp_medication_column] = (
+            pop_medicated[self.sbp_medication_column] + 1
+        )
+
+        return pop_medicated
+
+    def apply_sbp_treatment_ramp(self, pop_visitors: pd.DataFrame) -> pd.DataFrame:
```
rmudambi: Nit: code is easier to read if you define a function closer to its actual usage. In this case it would be easier to read if `apply_sbp_treatment_ramp` was defined above the `treat_...` methods

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+        """Applies the SBP treatment ramp
+
+        Arguments:
+            pop_visitors: dataframe subset to simulants visiting the doctor
+        """
+        currently_medicated = pop_visitors[
+            pop_visitors[self.sbp_medication_column].notna()
```
rmudambi: See above about using `nan` in this column

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -0,0 +1,361 @@
...
+        self.population_view.update(
+            pop[
+                [
+                    self.sbp_medication_column,
+                    self.ldlc_medication_column,
+                ]
+            ]
+        )
+
+    def treat_not_currently_medicated_sbp(
```
rmudambi: There is a lot of code duplication between currently medicated and newly medicated. I think splitting them into separate functions at this level is inappropriate. If we want to create new functions for different paths (which I'm not completely convinced is necessary), I would do it only once we get to the treatment assignment nodes (i.e. the ones labeled B, C, and D). If you make rung 0 on the ladder `NO_TREATMENT`, nodes B and D look identical to me.

src/vivarium_nih_us_cvd/utilities.py
```diff
@@ -176,3 +176,11 @@ def get_random_variable(draw: int, seeded_distribution: SeededDistribution) -> f
     seed, distribution = seeded_distribution
     np.random.seed(get_hash(f"{seed}_draw_{draw}"))
     return distribution.rvs()
+
+
+def get_measurement_error(
+    index: pd.Index, mean: float, sd: float, randomness: "RandomnessStream"
```
rmudambi: You need an `additional_key` argument here

rmudambi (changes requested):

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: Oof, this is contrived and I'd intended on re-factoring this but just kinda forgot. This actually IS a numerical column but it's pretty tricky to see how without looking closely. This column should at the very least probably be re-named to soemthing like "sbp_medication_ramp_level"

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: I agree, but this is black's doing

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: I'm aware :) - I needed the index anyway and just arbitrarily chose to use it here

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: Good catch

src/vivarium_nih_us_cvd/components/healthcare_visits.py
stevebachmeier: I think I understand but let's chat about this

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: iirc, adding it to the `columns_required` list didn't seem to make sense b/c it's not really "required" for `on_initialize_simulants`. Is there a cleaner way to do this?

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: I knew there was a chance that adherence could become more complicated; are you implying that the polypill interventions actually affect adherence?

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: We can leave this as is, but add a 0 level for "NO_TREATMENT"

mattkappel:
looks like you may need to isort and black

src/vivarium_nih_us_cvd/components/healthcare_visits.py
```diff
@@ -268,5 +276,8 @@ def random_time_delta(self, start: pd.Series, end: pd.Series) -> pd.Series:
         """Generate a random time delta for each individual in the start
         and end series."""
         return pd.to_timedelta(
-            start + (end - start) * self.randomness.get_draw(start.index), unit="day"
+            start
+            + (end - start)
+            * self.randomness.get_draw(start.index, additional_key="schedule_followup"),
```
rmudambi: The additional key should be defined in `schedule followup` and passed as an argument to `random_time_delta` in case you use the latter function again for some reason. Its name doesn't indicate that it will only be used for scheduling follow-up appointments.

Note: if `random_time_delta` is in fact not being used anywhere else, there's no real reason for it to exist as an independent function - it should just be moved inside `schedule_followup`.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -25,200 +32,186 @@ def __repr__(self):
...
         # Initialize medication coverage
-        pop[self.sbp_medication_column] = np.nan
-        pop[self.ldlc_medication_column] = np.nan
-        p_medication = self.calculate_medication_coverage_probabilities(pop)
+        pop[
+            data_values.COLUMNS.SBP_MEDICATION
+        ] = data_values.SBP_MEDICATION_LEVEL.NO_TREATMENT.DESCRIPTION
+        pop[
+            data_values.COLUMNS.LDLC_MEDICATION
+        ] = data_values.SBP_MEDICATION_LEVEL.NO_TREATMENT.DESCRIPTION
```
rmudambi: I know it's the same string, but shouldn't this be `LDLC_MEDICATION_LEVEL.NO_TREATMENT.DESCRIPTION`?

rmudambi (approved):
Looks good! Two minor nits, but you can merge once you look at them

PR #36, opened 2022-10-20
Feature/sbachmei/mic 3515 remove sbp treatment effect from gbd data

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3515](https://jira.ihme.washington.edu/browse/MIC-XYZ)
- *Research reference*: [creation of untreated sbp values on initialization](https://vivarium-research.readthedocs.io/en/latest/concept_models/vivarium_us_cvd/concept_model.html#creation-of-untreated-sbp-values-on-initializaiton)

GBD data includes inherent effects of SBP treatment. This estimates the
removal of those effects at the simulant level by increasing the gbd
values via an immutable multiplier that is determine from the
baseline medication coverage upon initialization.

### Verification and Testing
Confirmed that self.sbp is indeed self.gbd_sbp * multiplier.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/risks.py
+++ b/src/vivarium_nih_us_cvd/components/risks.py
@@ -30,7 +30,6 @@ class Risk(Risk_):
         return exposures
 
 
-# class SBPRisk(Risk_):
 class SBPRisk:
     """Manages gbd SBP exposure and untreated SBP exposure pipelines"""
 
--- a/src/vivarium_nih_us_cvd/components/risks.py
+++ b/src/vivarium_nih_us_cvd/components/risks.py
@@ -4,15 +4,47 @@ import pandas as pd
 from vivarium_public_health.risks.base_risk import Risk as Risk_
 
 from vivarium_nih_us_cvd.constants.data_values import RISK_EXPOSURE_LIMITS
+from vivarium_public_health.risks.data_transformations import get_exposure_post_processor
 
 
-# FIXME: would be more proper to modify the distribution function itself
 class Risk(Risk_):
     """Use the standard vivarium_public_health Risk class for risk
     exposure except apply limits to lower and upper bounds (when defined).
     """
 
     def _get_current_exposure(self, index: pd.Index) -> pd.Series:
+        exposures = super()._get_current_exposure(index)
+        if self.risk.name in RISK_EXPOSURE_LIMITS:
+            min_exposure = RISK_EXPOSURE_LIMITS[self.risk.name]["minimum"]
+            max_exposure = RISK_EXPOSURE_LIMITS[self.risk.name]["maximum"]
+            exposures[exposures < min_exposure] = min_exposure
+            exposures[exposures > max_exposure] = max_exposure
+        return exposures
+
+class SBPRisk(Risk_):
+    
+    def __init__(self, risk):
+        super().__init__(risk)
+
+    def setup(self, builder: 'Builder'):
+        # propensity_col = f'{self.risk.name}_propensity'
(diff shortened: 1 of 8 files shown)


src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -91,7 +97,30 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
...
+        mask_sbp_two_drugs = (
+            pop[data_values.COLUMNS.SBP_MEDICATION]
+            == data_values.SBP_MEDICATION_LEVEL.TWO_DRUGS_HALF_DOSE.DESCRIPTION
+        )
+        pop.loc[
+            (mask_sbp_adherent) & (mask_sbp_one_drug), data_values.COLUMNS.SBP_MULTIPLIER
+        ] = data_values.SBP_MULTIPLIER.ONE_DRUG
+        pop.loc[
+            (mask_sbp_adherent) & (mask_sbp_two_drugs), data_values.COLUMNS.SBP_MULTIPLIER
+        ] = data_values.SBP_MULTIPLIER.TWO_DRUGS
```
albrja, line 119: I like having the column name in data values.  It makes it easier to reference instead of having to type the string every time.

albrja (approved):

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: Yeah, i don't think data_values.py is really where it belongs, but def containerizing column names is a useful thing

src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -20,3 +28,70 @@ def _get_current_exposure(self, index: pd.Index) -> pd.Series:
...
+        return f"Risk({self.risk})"
+
+    #################
+    # Setup methods #
+    #################
+
+    # noinspection PyAttributeOutsideInit
+    def setup(self, builder: Builder) -> None:
+        self.gbd_exposure = self._get_gbd_exposure_pipeline(builder)
+        super().setup(builder)
```
rmudambi, line 49: I think we established the order doesn't actually matter here, so I'd suggest switching it back as it looks nicer to have  the `super` call first. If it does in fact need to be in this order, disregard this comment.

src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -20,3 +28,70 @@ def _get_current_exposure(self, index: pd.Index) -> pd.Series:
...
+            requires_values=[self.propensity_pipeline_name],
+            preferred_post_processor=get_exposure_post_processor(builder, self.risk),
+        )
+
+    def _get_exposure_pipeline(self, builder: Builder) -> Pipeline:
+        return builder.value.register_value_producer(
+            self.exposure_pipeline_name,
+            source=self._get_current_exposure,
+            requires_columns=[COLUMNS.SBP_MULTIPLIER],
+            requires_values=[self._get_gbd_exposure_pipeline],
```
rmudambi: Note that we need the pipeline NAME here, not the pipeline itself.

src/vivarium_nih_us_cvd/constants/data_keys.py
```diff
@@ -14,6 +14,11 @@ class SourceTarget(NamedTuple):
     target: str
 
 
+##########
```
rmudambi: Nit: Population isn't a cause :)

rmudambi (approved):

PR #37, opened 2022-10-21
Feature/sbachmei/mic 3516 apply treatment effect

stevebachmeier:
### Description
- *Category*: Implementation
- *JIRA issue*: [MIC-3516](https://jira.ihme.washington.edu/browse/MIC-3516)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/concept_models/vivarium_us_cvd/concept_model.html#treatment-effects

This PR implements SBP medication effects on sbp exposure values. The research team
provided a table that maps (binned) SBP values and medication level to an amount
of SBP to be subtracted. This is implemented as a value modifer to the sbp exposure
pipeline.

### Verification and Testing
Spot-checked that it's behaving as expected.


setup.py
```diff
@@ -49,6 +49,7 @@
         package_dir={"": "src"},
         packages=find_packages(where="src"),
         include_package_data=True,
+        package_data={"vivarium_nih_us_csv": ["data/drug_efficiency_sbp_new.csv"]},
```
rmudambi: Why are you doing this?

src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -1,18 +1,43 @@
...
+from vivarium_nih_us_cvd.constants.data_values import (
+    COLUMNS,
+    MEDICATION_ADHERENCE_SCORE,
+    RISK_EXPOSURE_LIMITS,
+    SBP_MEDICATION_LEVEL,
+)
+from vivarium_nih_us_cvd.constants.paths import FILEPATHS
+
+# Format the SBP risk effects file and generate bin edges
+sbp_risk_effects = pd.read_csv(FILEPATHS.SBP_MEDICATION_EFFECTS)
```
rmudambi: I think this should be done somewhere else, so that when you look at the csv file that exists in the repository it is the same as the data that is actually being used by the simulation. Isn't this a very small csv? Why can't you just re-format it manually?

I'm not a huge fan of code just living free in a file like this in general.

src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -44,10 +69,11 @@ def __repr__(self) -> str:
     # Setup methods #
     #################
 
-    # noinspection PyAttributeOutsideInit
     def setup(self, builder: Builder) -> None:
         super().setup(builder)
         self.gbd_exposure = self._get_gbd_exposure_pipeline(builder)
+        self.target_modifier = self._get_target_modifier(builder)
+        self._register_target_modifier(builder)
```
rmudambi: This target modifier should not be defined in the Risk component. It should be defined in the Treatment component

src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -95,3 +126,55 @@ def _get_current_exposure(self, index: pd.Index) -> pd.Series:
...
+
+    def _get_target_modifier(
+        self, builder: Builder
+    ) -> Callable[[pd.Index, pd.Series], pd.Series]:
+        """Apply medication effects"""
+
+        def adjust_target(index: pd.Index, target: pd.Series) -> pd.Series:
+            """Determine the exposure decrease as treatment_efficacy * adherence_score"""
+            pop_view = self.population_view.get(index)
+            adherence_scores = pop_view[COLUMNS.SBP_MEDICATION_ADHERENCE].map(
```
rmudambi: This can just be a boolean mask `pop_view[COLUMNS.SBP_MEDICATION_ADHERENCE] = MEDICATION_ADHERENCE_TYPE.ADHERENT`. 
```
int x = 4
x * False == 0
x * True == 4
```

src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -95,3 +126,55 @@ def _get_current_exposure(self, index: pd.Index) -> pd.Series:
...
+                MEDICATION_ADHERENCE_SCORE
+            )
+            df_efficacy = pd.DataFrame(
+                {"bin": pd.cut(x=target, bins=sbp_bin_edges, right=True)}
+            )
+            df_efficacy["sbp_start_exclusive"] = df_efficacy["bin"].apply(lambda x: x.left)
```
rmudambi: I'd like to talk about this implementation. I think there's a simpler way to do this, but it'd be easier to work through it together

src/vivarium_nih_us_cvd/constants/data_values.py
```diff
@@ -247,19 +247,12 @@ def name(self):
...
-    PRIMARY_NON_ADHERENT: float = 0.0
-    SECONDARY_NON_ADHERENT: float = 0.0
-
-    @property
-    def name(self):
-        return "medication_adherence_score"
-
-
-MEDICATION_ADHERENCE_SCORE = __MedicationAdherenceScore()
+MEDICATION_ADHERENCE_SCORE = {
```
rmudambi: This is not needed

rmudambi (changes requested):

albrja (approved):

setup.py
stevebachmeier: So that the file exists with future installs.

src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -95,3 +126,55 @@ def _get_current_exposure(self, index: pd.Index) -> pd.Series:
...
+            # Simulants not on treatment mean 0 effect
+            assert set(
+                df_efficacy.loc[df_efficacy["value"].isna(), COLUMNS.SBP_MEDICATION]
+            ) == {SBP_MEDICATION_LEVEL.NO_TREATMENT.DESCRIPTION}
+            df_efficacy.loc[
+                df_efficacy[COLUMNS.SBP_MEDICATION]
+                == SBP_MEDICATION_LEVEL.NO_TREATMENT.DESCRIPTION,
+                "value",
+            ] = 0
+            assert df_efficacy["value"].isna().sum() == 0
```
albrja: Would these asserts ever be raised?

stevebachmeier: Not sure I understand the question. If the assertion fails it had better break b/c something went wrong

albrja: Is this just a check for sanity sake?  The assertion will break if something goes wrong but shouldn't you handle this when you are processing before you hit the assert statement?    This is a good check but it will just break and you will have to go back and find the source of the issue if you are expected these actually be hit.

albrja: In other words, disregard this comment if you aren't expecting to to raise the assert statement but just added them to be safe.

src/vivarium_nih_us_cvd/components/risks.py
stevebachmeier: I reformatted it as requested, though I still had to do a bit of wrangling to get the lower and upper bounds for cutting (-inf and +inf). Happy to discuss a better approach - there's gotta be one!

(I considered just merging directly on the bin label, eg "(125, 135]", just I really dislike working with strings)

src/vivarium_nih_us_cvd/components/risks.py
stevebachmeier: Yeah I guess I was trying to keep it flexible to non-binary adherence scores (but admittedly at the expense of simplicity)

src/vivarium_nih_us_cvd/components/risks.py
stevebachmeier: Happy to chat about if you decide it's a priority

hussain-jafari (approved):

PR #41, opened 2022-10-26
implement exposure to ldlc medication

stevebachmeier:
### Description
- *Category*: Implementation
- *JIRA issue*: [MIC-3375](https://jira.ihme.washington.edu/browse/MIC-3375)
- *Research reference*: [system modeling](https://vivarium-research.readthedocs.io/en/latest/concept_models/vivarium_us_cvd/concept_model.html#healthcare-system-modeling) (refer to the LDL-C Treatment Ramp)

This PR exposes simulants to the ldlc-treatment ramp and moves them up/down
the medication ramp as appropriate. Note that baseline ldlc medication coverage
already exists from a previous PR (ie, simulants were already initialized
with ldlc medication levels); but now they can go from a no treatment
state to a treatment state as well as move up the ramp as appropriate.

Also note that this does not yet implement the effect of the ldlc medication.

The majority of the changes lie in the self.apply_ldlc_treatment_ramp() method.
If helpful, refer to my proposed design here to see where the
logic is coming from: https://uwnetid-my.sharepoint.com/:u:/g/personal/sbachmei_uw_edu/EdHAaQc2PppPt2HumYvBeiABQpMhiIfgOQjnAqWGdXF77w?e=2m7RqH

### Verification and Testing
Ran interactively and ensured simulants are moving up the ldlc treatment ramp

### DIFF
--- a/src/vivarium_nih_us_cvd/components/treatment.py
+++ b/src/vivarium_nih_us_cvd/components/treatment.py
@@ -590,13 +590,14 @@ class Treatment:
         """Introduce a measurement error to the sbp exposure values"""
         if not exposure_pipeline:
             exposure_pipeline = self.sbp
-        return exposure_pipeline(index) + get_random_value_from_normal_distribution(
+        
+        return (exposure_pipeline(index) + get_random_value_from_normal_distribution(
             index=index,
             mean=mean,
             sd=sd,
             randomness=self.randomness,
             additional_key="measured_sbp",
-        )
+        )).clip(lower=0)
 
     def get_ascvd(
         self, pop_visitors: pd.DataFrame, sbp_pipeline: Optional[Pipeline] = None
@@ -625,10 +626,11 @@ class Treatment:
         """Introduce a measurement error to the ldlc exposure values"""
         if not exposure_pipeline:
             exposure_pipeline = self.ldlc
-        return exposure_pipeline(index) + get_random_value_from_normal_distribution(
+        
+        return (exposure_pipeline(index) + get_random_value_from_normal_distribution(
             index=index,
             mean=mean,
             sd=sd,
             randomness=self.randomness,
             additional_key="measured_ldlc",
-        )
+        )).clip(lower=0)
--- a/src/vivarium_nih_us_cvd/components/treatment.py
+++ b/src/vivarium_nih_us_cvd/components/treatment.py
@@ -490,8 +490,7 @@ class Treatment:
         )
         history_mi_or_is = pop_visitors[mask_history_mi | mask_history_is].index
         newly_prescribed = (
-            overcome_therapeutic_inertia
-            .difference(currently_medicated)
+            overcome_therapeutic_inertia.difference(currently_medicated)
             .difference(low_ascvd)
             .difference(low_ldlc)
         )
@@ -499,36 +498,29 @@ class Treatment:
         # [Treatment ramp ID D] Simulants who overcome therapeutic inertia, have
         # elevated LDLC, are not currently medicated, have elevated ASCVD, and
         # have a history of MI or IS
-        to_prescribe_d = (
-            newly_prescribed
-            .intersection(history_mi_or_is)
-        )
+        to_prescribe_d = newly_prescribed.intersection(history_mi_or_is)
         # [Treatment ramp ID E] Simulants who overcome therapeutic inertia, have
         # elevated LDLC, are not currently medicated, have elevated ASCVD, have
         # no history of MI or IS, and who have high LDLC or ASCVD
-        to_prescribe_e = (
-            newly_prescribed
-            .difference(to_prescribe_d)
-            .intersection(high_ascvd.union(high_ldlc))
+        to_prescribe_e = newly_prescribed.difference(to_prescribe_d).intersection(
+            high_ascvd.union(high_ldlc)
         )
         # [Treatment ramp ID F] Simulants who overcome therapeutic inertia, have
         # elevated LDLC, are not currently medicated, have elevated ASCVD, have
         # no history of MI or IS, but who do NOT have high LDLC or ASCVD
-        to_prescribe_f = (
-            newly_prescribed
-            .difference(to_prescribe_d)
-            .difference(to_prescribe_e)
+        to_prescribe_f = newly_prescribed.difference(to_prescribe_d).difference(
+            to_prescribe_e
         )
         # [Treatment ramp ID G] Simulants who overcome therapeutic inertia, have
         # elevated LDLC, and are currently medicated
-        to_prescribe_g = (
-            overcome_therapeutic_inertia
-            .intersection(currently_medicated)
-            .difference(low_ldlc)
-        )
+        to_prescribe_g = overcome_therapeutic_inertia.intersection(
+            currently_medicated
+        ).difference(low_ldlc)
 
         # Prescribe initial medications
-        df_newly_prescribed = pd.DataFrame(index=to_prescribe_e.union(to_prescribe_f).union(to_prescribe_d))
+        df_newly_prescribed = pd.DataFrame(
+            index=to_prescribe_e.union(to_prescribe_f).union(to_prescribe_d)
+        )
         # TODO: CONFIRM THESE LISTS REMAIN ORDERED HERE AND IN randomness.choice
         df_newly_prescribed.loc[
             to_prescribe_d,
--- a/src/vivarium_nih_us_cvd/components/treatment.py
+++ b/src/vivarium_nih_us_cvd/components/treatment.py
@@ -382,8 +382,6 @@ class Treatment:
 
         measured_sbp = self.get_measured_sbp(
             index=pop_visitors.index,
-            mean=data_values.MEASUREMENT_ERROR_MEAN_SBP,
-            sd=data_values.MEASUREMENT_ERROR_SD_SBP,
             exposure_pipeline=exposure_pipeline,
         )
 
@@ -475,8 +473,6 @@ class Treatment:
         ascvd = self.get_ascvd(pop_visitors=pop_visitors, sbp_pipeline=sbp_pipeline)
         measured_ldlc = self.get_measured_ldlc(
             index=pop_visitors.index,
-            mean=data_values.MEASUREMENT_ERROR_MEAN_LDLC,
-            sd=data_values.MEASUREMENT_ERROR_SD_LDLC,
             exposure_pipeline=ldlc_pipeline,
         )
 
@@ -583,8 +579,6 @@ class Treatment:
     def get_measured_sbp(
         self,
         index: pd.Index,
-        mean: float,
-        sd: float,
         exposure_pipeline: Optional[Pipeline] = None,
     ) -> pd.Series:
         """Introduce a measurement error to the sbp exposure values"""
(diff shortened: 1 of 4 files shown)


mattkappel:
I'm not able to get at that diagram. Maybe you could drag and drop a render of it into the PR description?

src/vivarium_nih_us_cvd/constants/data_values.py
```diff
@@ -101,10 +101,14 @@ def name(self):
 # Medication Parameters #
 #########################
 
-THERAPEUTIC_INERTIA_NO_START = (
+SBP_THERAPEUTIC_INERTIA_NO_START = (
```
rmudambi: What does "no start" mean? I think just call it `SBP_THERAPEUTIC_INERTIA`

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -363,15 +365,14 @@ def apply_sbp_treatment_ramp(
...
-
         if not exposure_pipeline:
             exposure_pipeline = self.sbp
         overcome_therapeutic_inertia = pop_visitors[
             self.randomness.get_draw(
                 pop_visitors.index,
                 additional_key="sbp_therapeutic_inertia",
             )
-            > data_values.THERAPEUTIC_INERTIA_NO_START
+            > data_values.SBP_THERAPEUTIC_INERTIA_NO_START
```
rmudambi: I couldn't find in the documentation any information about whether a given individual's therapeutic inertia is static or should be recalculated each on each doctor's visit. Can you show me where that is stated? If it's not clear there, make sure you get that clarification from the RT. 

This implementation assumes there is a new draw at each visit. If it's a static value this will need to be changed and will also have implications for the current approach to initialization.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -429,15 +430,154 @@ def apply_sbp_treatment_ramp(
...
+        currently_medicated = pop_visitors[
+            pop_visitors[data_values.COLUMNS.LDLC_MEDICATION]
+            != data_values.LDLC_MEDICATION_LEVEL.NO_TREATMENT.DESCRIPTION
+        ].index
+        not_currently_medicated = pop_visitors.index.difference(currently_medicated)
+
+        ascvd = self.get_ascvd(pop_visitors=pop_visitors, sbp_pipeline=sbp_pipeline)
+        measured_ldlc = self.get_measured_ldlc(
+            index=pop_visitors.index,
+            mean=data_values.MEASUREMENT_ERROR_MEAN_LDLC,
```
rmudambi: Mean and SD shouldn't be parameters to this function since they are always the same.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -429,15 +430,154 @@ def apply_sbp_treatment_ramp(
...
+            sd=data_values.MEASUREMENT_ERROR_SD_LDLC,
+            exposure_pipeline=ldlc_pipeline,
+        )
+
+        # Generate other useful helper indexes
+        elevated_ascvd = ascvd[ascvd >= data_values.ASCVD_THRESHOLD.LOW].index
```
rmudambi: It would be less confusing just to have low, medium, and high indexes for ASCVD and LDL-C where `low == ~elevated` and you use the difference operator rather than intersection.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -429,15 +430,154 @@ def apply_sbp_treatment_ramp(
...
+        mask_history_mi = (
+            pop_visitors[models.MYOCARDIAL_INFARCTION_MODEL_NAME]
+            != models.MYOCARDIAL_INFARCTION_SUSCEPTIBLE_STATE_NAME
+        )
+        mask_history_is = (
+            pop_visitors[models.ISCHEMIC_STROKE_MODEL_NAME]
+            != models.ISCHEMIC_STROKE_SUSCEPTIBLE_STATE_NAME
+        )
+        history_mi_or_is = pop_visitors[mask_history_mi | mask_history_is].index
+        no_history_mi_or_is = pop_visitors.index.difference(history_mi_or_is)
```
rmudambi: When you have these binary options you don't need to create both indexes. You can just choose to use `intersection` or `difference` as needed for a single index. 

For example, below you could do 
```
to_prescrbe_e = (
    overcome_therapeutic_inertia
    .difference(low_ldlc)
    .difference(currently_medicated)
    .difference(low_ascvd)
    .difference(history_mi_or_is)
    .intersection(high_ascvd).union(high_ldlc))
)
```

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -429,15 +430,154 @@ def apply_sbp_treatment_ramp(
...
+
+        # [Treatment ramp ID E] Simulants who overcome therapeutic inertia, have
```
rmudambi, line 502: To reduce the complexity of all the index arithmetic below you can instead create a single index like so:
```
newly_prescribed = (
    overcome_therapeutic_inertia
    .difference(low_ldlc)
    .difference(currently_medicated)
    .difference(low_ascvd)
)
```

Then you can simplify your later formulations:
```
to_prescribe_d = (
    newly_prescribed
    .intersection(history_mi_or_is)
)
to_prescribe_e = (
    newly_prescribed
    .difference(to_prescribe_d)
    .intersection(high_ascvd).union(high_ldlc))
)
to_prescribe_f = (
    newly_prescribed
    .difference(to_prescribe_d)
    .difference(to_prescribe_e)
)
```

Or if you don't like the references to `to_prescribe_d/e` you can do:
```
to_prescribe_e = (
    newly_prescribed
    .difference(history_mi_or_is)
    .intersection(high_ascvd).union(high_ldlc))
)
to_prescribe_f = (
    newly_prescribed
    .difference(history_mi_or_is)
    .difference(high_ascvd)
    .difference(high_ldlc)
)
```

I think I prefer the former formulation over the latter though it does obfuscate the rules a little, since it guarantees that E, F,  and G will be mutually exclusive and collectively exhaustive. It's easy to make a mistake here and get overlaps or gaps (as you actually do below).

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -429,15 +430,154 @@ def apply_sbp_treatment_ramp(
...
+        )
+        # [Treatment ramp ID F] Simulants who overcome therapeutic inertia, have
+        # elevated LDLC, are not currently medicated, have elevated ASCVD, have
+        # no history of MI or IS, but who do NOT have high LDLC or ASCVD
+        to_prescribe_f = (
+            overcome_therapeutic_inertia.intersection(elevated_ldlc)
+            .intersection(not_currently_medicated)
+            .intersection(elevated_ascvd)
+            .intersection(no_history_mi_or_is)
+            .intersection(mid_ascvd.union(mid_ldlc))
```
rmudambi: This is actually incorrect, as it will include simulants who have one (but not both) of high ASCVD and high LDL-C, when we really want them both to be mid.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -429,15 +430,154 @@ def apply_sbp_treatment_ramp(
...
+        # have a history of MI or IS
+        to_prescribe_d = (
+            overcome_therapeutic_inertia.intersection(elevated_ldlc)
+            .intersection(not_currently_medicated)
+            .intersection(elevated_ascvd)
+            .intersection(history_mi_or_is)
+        )
+        # [Treatment ramp ID G] Simulants who overcome therapeutic inertia, have
+        # elevated LDLC, and are currently medicated
+        to_prescribe_g = overcome_therapeutic_inertia.intersection(
```
rmudambi, line 516: The style you have above with each operation on a new line is much easier to read than this.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -450,24 +590,29 @@ def get_measured_sbp(
         """Introduce a measurement error to the sbp exposure values"""
         if not exposure_pipeline:
             exposure_pipeline = self.sbp
-        return exposure_pipeline(index) + get_random_value_from_normal_distribution(
+        
+        return (exposure_pipeline(index) + get_random_value_from_normal_distribution(
```
rmudambi: This doesn't seem black-adherent?

rmudambi (changes requested):

stevebachmeier:
> 

@mattkappel Try the updated link (I hate permissions in google drive)

src/vivarium_nih_us_cvd/constants/data_values.py
stevebachmeier: It was to help me remember that that's the percentage of people that won't start. No issue changing it, though.

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: Oh good point. I feel like we've discussed this but I'm not seeing it either. I'll confirm.

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: Ah, good catch. At one point they werre the same function but now that I had to dork around with the exposure pipelines as well I split them and forgot to consolidate.

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: That's a black thing - I'll try and force it

stevebachmeier: Nope, black won't let me

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: understood. It's easier for me to keep the logical (ie `.intersection`) consistent if possible. Though sometimes it does come at the expense of requiring more variables here. I'll have a look and see if I can simplify in a way I'm comfortable wtih

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: lol I'd actually done this already but didn't love the reference to previous indexes. BUT, happy to change it back if it's more readable to others (ie you); especially since I'd literally done that initially

src/vivarium_nih_us_cvd/components/treatment.py
rmudambi: If you wrap this in parentheses it will

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: oof, good catch. Yeah I'm now convinced that your recommended method is better.

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: I tried. black removed them :(. I'll confirm though

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: Syl confirmed to draw against inertia at every visit and will update the docs.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -429,62 +428,196 @@ def apply_sbp_treatment_ramp(
...
+            to_prescribe_e
+        )
+        # [Treatment ramp ID G] Simulants who overcome therapeutic inertia, have
+        # elevated LDLC, and are currently medicated
+        to_prescribe_g = overcome_therapeutic_inertia.intersection(
+            currently_medicated
+        ).difference(low_ldlc)
+
+        # Prescribe initial medications
+        df_newly_prescribed = pd.DataFrame(
```
rmudambi: No need for this union.

rmudambi (approved):

mattkappel (approved):

PR #50, opened 2022-11-08
Feature/sbachmei/mic 3556 outreach effect

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3556](https://jira.ihme.washington.edu/browse/MIC-3556)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/intervention_models/crm_mgmt/vivarium_outreach.html#vivarium-modeling-strategy

This PR implements the outreach effects on adherence values. The approach:
- register a value modifier in the OutreachEffect component that
  modifies the adherence pipelines (the outreach program essentially increases
  people's medication adherence levels)
- In the Treatment component, update the 'outreach' state table column
  by setting it equal to the outreach pipeline _only for simulants where the
  current pipeline is different than the column_ (ie new outreaches only)
- Then, update the adherence columns for those being enrolled
  by setting the columns equal to the adherence pipelines. The changes
  will happen due to the just-updated 'outreach' column

### Verification and Testing
I've made sure the code runs, but it's currently difficult to test
because without the outreach exposure changing (next PR
when I implement scale-up), there are never any new people
to enroll.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/effects.py
+++ b/src/vivarium_nih_us_cvd/components/effects.py
@@ -92,12 +92,9 @@ class OutreachEffect:
         builder.value.register_value_modifier(
             self.sbp_target_pipeline_name,
             modifier=self.sbp_medication_adherence_target_modifier,
-            # requires_columns=self.columns_required,
-            # requires_values=[],
         )
 
         builder.value.register_value_modifier(
             self.ldlc_target_pipeline_name,
             modifier=self.ldlc_medication_adherence_target_modifier,
-            # requires_columns=self.columns_required,
         )
--- a/src/vivarium_nih_us_cvd/components/effects.py
+++ b/src/vivarium_nih_us_cvd/components/effects.py
@@ -47,19 +47,25 @@ class OutreachEffect:
             modifier=self._ldlc_adherence_modifier(builder),
         )
 
-    def _sbp_adherence_modifier(self, builder: Builder) -> Callable[[pd.Index, pd.Series], pd.Series]:
+    def _sbp_adherence_modifier(
+        self, builder: Builder
+    ) -> Callable[[pd.Index, pd.Series], pd.Series]:
         def adjust_target(index: pd.Index, target: pd.Series) -> pd.Series:
             return self._adjust_target(index=index, target=target, medication_type="sbp")
-        
+
         return adjust_target
 
-    def _ldlc_adherence_modifier(self, builder: Builder) -> Callable[[pd.Index, pd.Series], pd.Series]:
+    def _ldlc_adherence_modifier(
+        self, builder: Builder
+    ) -> Callable[[pd.Index, pd.Series], pd.Series]:
         def adjust_target(index: pd.Index, target: pd.Series) -> pd.Series:
             return self._adjust_target(index=index, target=target, medication_type="ldlc")
-        
+
         return adjust_target
 
-    def _adjust_target(self, index: pd.Index, target: pd.Series, medication_type: str) -> pd.Series:
+    def _adjust_target(
+        self, index: pd.Index, target: pd.Series, medication_type: str
+    ) -> pd.Series:
         clock_time = self.clock()
         breakpoint()  # save for next PR
         # Do not adjust target on intialization. We do this because during
--- a/src/vivarium_nih_us_cvd/components/effects.py
+++ b/src/vivarium_nih_us_cvd/components/effects.py
@@ -39,12 +39,12 @@ class OutreachEffect:
     def _register_target_modifiers(self, builder: Builder) -> None:
         builder.value.register_value_modifier(
             data_values.PIPELINES.SBP_MEDICATION_ADHERENCE_EXPOSURE,
-            modifier=self._sbp_adherence_modifier,
+            modifier=self._sbp_adherence_modifier(builder),
         )
 
         builder.value.register_value_modifier(
             data_values.PIPELINES.LDLC_MEDICATION_ADHERENCE_EXPOSURE,
-            modifier=self._ldlc_adherence_modifier,
+            modifier=self._ldlc_adherence_modifier(builder),
         )
 
     def _sbp_adherence_modifier(self, builder: Builder) -> Callable[[pd.Index, pd.Series], pd.Series]:
--- a/src/vivarium_nih_us_cvd/components/effects.py
+++ b/src/vivarium_nih_us_cvd/components/effects.py
@@ -12,13 +12,7 @@ class OutreachEffect:
     """A component to model the impact of the outreach risk on medication adherence levels"""
 
     def __init__(self):
-        self.exposure_pipeline_name = data_values.PIPELINES.OUTREACH_EXPOSURE
-        self.sbp_target_pipeline_name = (
-            data_values.PIPELINES.SBP_MEDICATION_ADHERENCE_EXPOSURE
-        )
-        self.ldlc_target_pipeline_name = (
-            data_values.PIPELINES.LDLC_MEDICATION_ADHERENCE_EXPOSURE
-        )
+        pass
 
     def __repr__(self):
         return "OutreachEffect"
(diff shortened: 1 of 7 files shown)


src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -0,0 +1,100 @@
...
+from vivarium.framework.time import get_time_stamp
+
+from vivarium_nih_us_cvd.constants import data_values
+
+
+class OutreachEffect:
+    """A component to model the impact of the outreach risk on medication adherence levels"""
+
+    def __init__(self):
+        self.exposure_pipeline_name = data_values.PIPELINES.OUTREACH_EXPOSURE
```
rmudambi: No need to define these constants again. They're already constants defined in `data_values`

src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -0,0 +1,100 @@
...
+    def name(self) -> str:
+        return f"outreach_effect"
+
+    #################
+    # Setup methods #
+    #################
+
+    def setup(self, builder: Builder) -> None:
+        self.randomness = builder.randomness.get_stream(self.name)
+        self.population_view = builder.population.get_view(["age", "sex"])
```
rmudambi: Why do you need these columns? I don't see them used anywhere

src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -0,0 +1,100 @@
...
+                additional_key=f"adjust_{medication_type}_medication_adherence_values",
+            )
+            target[primary_non_adherent] = adjusted
+
+        return target
+
+    def _register_target_modifiers(self, builder: Builder) -> None:
+        builder.value.register_value_modifier(
+            self.sbp_target_pipeline_name,
+            modifier=self.sbp_medication_adherence_target_modifier,
```
rmudambi: I think this layer of indirection is unneeded. You can just say `modifier=self._sbp_adherence_modifier`, where `self._sbp_adherence_modifier` is the `adjust_target` function inside `_get_sbp_medication_adherence_target_modifier`. Same for LDL.

This has the twofold benefit of reducing boilerplate and making the code easier to read.

src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -0,0 +1,100 @@
...
+            return self._adjust_target(
+                index=index, target=target, medication_type="ldlc", builder=builder
+            )
+
+        return adjust_target
+
+    def _adjust_target(
+        self, index: pd.Index, target: pd.Series, medication_type: str, builder: Builder
+    ) -> pd.Series:
+        clock_time = builder.time.clock()()
```
rmudambi: Define `self.clock = builder.time.clock()` in the setup method and then you can just call `self.clock()` here.

src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -0,0 +1,100 @@
...
+        simulation_start_time = get_time_stamp(builder.configuration.time.start)
+        breakpoint()  # save for next PR
+        # Do not adjust target on intialization
```
rmudambi: Add a note about **why** we don't want to apply this on initialization.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -644,6 +659,55 @@ def apply_ldlc_treatment_ramp(
             + 1
         ).map(self.ldlc_treatment_map)
 
+        # Apply outreach intervention to groups 'newly_prescribed' (d, e, f)
+        # and simulants already on medication
+        maybe_enroll = newly_prescribed.union(currently_medicated)
+        pop_visitors = self.enroll_in_outreach(pop_visitors, maybe_enroll)
```
rmudambi: This shouldn't happen in the treatment ramp functions themselves. You should just mark these simulants to be maybe enrolled and then call `enroll_in_outreach` once outside the ramp functions. As it is, if a simulant is on both types of medication this function will be called on them twice in the same timestep.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -644,6 +659,55 @@ def apply_ldlc_treatment_ramp(
...
+                to_enroll,
+                [
+                    data_values.COLUMNS.SBP_MEDICATION_ADHERENCE,
+                    data_values.COLUMNS.LDLC_MEDICATION_ADHERENCE,
+                ],
+            ] = [
+                self.sbp_medication_adherence(to_enroll),
+                self.ldlc_medication_adherence(to_enroll),
+            ]
+            self.population_view.update(
```
rmudambi, line 723: Why are you updating these columns both here and in line 440?

src/vivarium_nih_us_cvd/constants/data_values.py
```diff
@@ -63,6 +64,7 @@ class __TimestepCleanupPriorities(NamedTuple):
 
     HEALTHCARE_VISITS: int = 5
     TREATMENT: int = 6
+    OUTREACH: int = 7
```
rmudambi: This is never used.

rmudambi (changes requested):

src/vivarium_nih_us_cvd/constants/data_values.py
stevebachmeier: Ah, lingering from when I tried haveing OutreachEffect initialize/manage the 'outreach' column.

src/vivarium_nih_us_cvd/components/effects.py
stevebachmeier: Oh yeah, I wanted to ask about that; it's no longer used since I'm not modifying the 'outreach' column here, but I was originally needing to load the state table and initialize the 'outreach' column. However, that initialization did not rely on any other columns - but it seems like you can't get a view of just the index?

src/vivarium_nih_us_cvd/components/treatment.py
stevebachmeier: oversight. For my own clarification, though - it wouldn't change anything right?

src/vivarium_nih_us_cvd/components/effects.py
stevebachmeier: Ooh, i like that

src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -12,13 +12,7 @@ class OutreachEffect:
...
 
     def __init__(self):
-        self.exposure_pipeline_name = data_values.PIPELINES.OUTREACH_EXPOSURE
-        self.sbp_target_pipeline_name = (
-            data_values.PIPELINES.SBP_MEDICATION_ADHERENCE_EXPOSURE
-        )
-        self.ldlc_target_pipeline_name = (
-            data_values.PIPELINES.LDLC_MEDICATION_ADHERENCE_EXPOSURE
-        )
+        pass
```
rmudambi: Why have an init that just passes? Just remove this

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -419,8 +443,21 @@ def on_time_step_cleanup(self, event: Event) -> None:
...
+        pop.loc[visitors], maybe_enroll_in_outreach_sbp = self.apply_sbp_treatment_ramp(
+            pop_visitors=pop.loc[visitors]
+        )
+        pop.loc[visitors], maybe_enroll_in_outreach_ldlc = self.apply_ldlc_treatment_ramp(
+            pop_visitors=pop.loc[visitors]
+        )
+
+        # Enroll in outreach intervention
+        breakpoint()
+        maybe_enroll_in_outreach = maybe_enroll_in_outreach_sbp.union(
```
mattkappel, line 455: probably don't want this

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -419,8 +443,21 @@ def on_time_step_cleanup(self, event: Event) -> None:
...
-        pop.loc[visitors] = self.apply_ldlc_treatment_ramp(pop_visitors=pop.loc[visitors])
+        pop.loc[visitors], maybe_enroll_in_outreach_sbp = self.apply_sbp_treatment_ramp(
+            pop_visitors=pop.loc[visitors]
+        )
+        pop.loc[visitors], maybe_enroll_in_outreach_ldlc = self.apply_ldlc_treatment_ramp(
+            pop_visitors=pop.loc[visitors]
+        )
+
+        # Enroll in outreach intervention
+        breakpoint()
```
mattkappel: probably don't want this

mattkappel (approved):
Looks good, save for the leftover breakpoint from debugging. I love the comments and PR description -- helpful for review. 👍

rmudambi (approved):

PR #51, opened 2022-11-09
Feature/sbachmei/mic 3557 outreach scaleup

stevebachmeier:
### Description
- *Category*: Implementation
- *JIRA issue*: [MIC-3557](https://jira.ihme.washington.edu/browse/MIC-3557)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/intervention_models/crm_mgmt/vivarium_outreach.html

### Verification and Testing
Ran a few timesteps and confirmed scale-up is happening.


src/vivarium_nih_us_cvd/components/interventions.py
```diff
@@ -0,0 +1,6 @@
+from vivarium_public_health.treatment import LinearScaleUp
+
+
+class OutreachInterventionScaleUp(LinearScaleUp):
```
rmudambi: Why are you setting this to None? It seems unnecessary and would enable you not to have to subclass at all.

rmudambi (approved):
Looks good, but unless there's a compelling reason that I'm missing, you don't need to subclass `LinearScaleUp` and shouldn't if you don't need to.

src/vivarium_nih_us_cvd/components/interventions.py
stevebachmeier: The LinearScaleUp defaults were causing the break so I just wipe them out.

https://github.com/ihmeuw/vivarium_public_health/blob/main/src/vivarium_public_health/treatment/scale_up.py#L142-L143

mattkappel (approved):

src/vivarium_nih_us_cvd/components/interventions.py
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    def _get_configuration_defaults(self) -> Dict[str, Dict]:
+        return {self.configuration_key: LinearScaleUp.configuration_defaults["treatment"]}
+
+    def _get_scale_up_dates(self, builder: Builder) -> Tuple[datetime, datetime]:
+        scale_up_config = builder.configuration[self.configuration_key]["date"]
+
+        def get_endpoint(endpoint_type: str) -> datetime:
```
rmudambi, line 38:
```
for endpoint_type in ["start", "end"]:
    method body
    ...
return ...
```

PR #55, opened 2022-11-15
indent scaleup values into intervention list

stevebachmeier:
### Description
- *Category*: bugfix
- *JIRA issue*: na
- *Research reference*: na

The model spec scaleup value was used for all non-baseline scenarios 
because the mappings did not exactly match between the two yamls.

### Verification and Testing
ran a small psimulate and confirmed scaling up correctly

### DIFF
--- a/src/vivarium_nih_us_cvd/components/interventions.py
+++ b/src/vivarium_nih_us_cvd/components/interventions.py
@@ -35,7 +35,7 @@ class LinearScaleUp(LinearScaleUp_):
         return {self.configuration_key: LinearScaleUp.configuration_defaults["treatment"]}
 
     def _get_scale_up_dates(self, builder: Builder) -> Tuple[datetime, datetime]:
-        scale_up_config = builder.configuration[self.configuration_key]["date"]
+        scale_up_config = builder.configuration.intervention[self.configuration_key]["date"]
         endpoints = {}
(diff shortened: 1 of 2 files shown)


src/vivarium_nih_us_cvd/components/interventions.py
```diff
@@ -35,7 +35,7 @@ def _get_configuration_defaults(self) -> Dict[str, Dict]:
         return {self.configuration_key: LinearScaleUp.configuration_defaults["treatment"]}
 
     def _get_scale_up_dates(self, builder: Builder) -> Tuple[datetime, datetime]:
-        scale_up_config = builder.configuration[self.configuration_key]["date"]
+        scale_up_config = builder.configuration.intervention[self.configuration_key]["date"]
```
stevebachmeier, line 38: This looks to the configuration.intervention key just like the super class (LinearScaleUp) does for determining whether the scenario is "baseline" or not

rmudambi, line 38: If you remove the indentation below, you won't need this line. If it doesn't work without it, something is wrong somewhere else.

rmudambi (changes requested):

stevebachmeier:
fixing on the branches file instead. closing this pr

PR #57, opened 2022-11-16
Feature/sbachmei/mic 3537 polypill implementation

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3537](https://jira.ihme.washington.edu/browse/MIC-3537)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/intervention_models/crm_mgmt/vivarium_polypill.html#intervention-crm-mgmt-polypill

The polypill intervention is very similar to the outreach intervention in that
it modifies medication adherence levels. Some notes:
* I had to refactor things a bit to make a single InterventionEffect class
  rather than different ones for each intervention
* Related, both interventions are mutually 
  exclusive (ie we will never run a polypill + outreach scenario). However,
  each of those scenarios modifes the sbp_medication_adherence
  pipeline so I needed to add some logic to register the correct
  one depending on the scenario.

### Verification and Testing
Ran some time steps and made sure adherence was changing


src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -67,8 +81,26 @@ def _adjust_target(
...
                 choices=list(data_values.OUTREACH_EFFECTS[medication_type].keys()),
                 p=list(data_values.OUTREACH_EFFECTS[medication_type].values()),
-                additional_key=f"adjust_{medication_type}_medication_adherence_values",
+                additional_key=f"outreach_adjust_{medication_type}_medication_adherence_values",
             )
             target[primary_non_adherent] = adjusted
 
         return target
+
+    def _polypill_sbp_adherence_modifier(
```
rmudambi, line 89: This layer of indirection is not needed since `_polypill_sbp_adherence_modifier` is identical to `_polypill_adjust_target`

src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -67,8 +81,26 @@ def _adjust_target(
...
+        self, index: pd.Index, target: pd.Series
+    ) -> pd.Series:
+        return self._polypill_adjust_target(index=index, target=target)
+
+    def _polypill_adjust_target(self, index: pd.Index, target: pd.Series) -> pd.Series:
+        polypill = self.polypill(index)
+        on_polypill = polypill[polypill == "cat1"].index
+        adjusted = target.copy()
```
rmudambi: Why do you need to copy it here?

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -331,13 +340,20 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
...
-        maybe_enroll_in_outreach = maybe_enroll_in_outreach_sbp.union(
-            maybe_enroll_in_outreach_ldlc
-        )
-        pop.loc[mask_emergency] = self.enroll_in_outreach(
-            pop_visitors=pop.loc[mask_emergency], maybe_enroll=maybe_enroll_in_outreach
-        )
+        # Enroll in interventions. The sbp treatment ramp includes both
+        # outreach and polypill enrollment logic whilc the ldlc ramp
+        # only includes outreach
+        if self.scenario.is_outreach_scenario:
```
rmudambi: Why would we be enrolling anyone in an intervention during initialization? Coverage should always be 0

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -552,16 +578,28 @@ def apply_sbp_treatment_ramp(
 
         # Determine potential new outreach enrollees; applies to groups b, c, and
         # everyone already on medication
-        maybe_enroll = to_prescribe_b.union(to_prescribe_c).union(currently_medicated)
+        maybe_enroll_outreach = to_prescribe_b.union(to_prescribe_c).union(
```
rmudambi: Should this be scenario specific? i.e. if we are not in an outreach scenario `maybe_enroll_outreach` should always be empty?

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -700,26 +738,54 @@ def enroll_in_outreach(
...
+    def enroll_in_polypill(
+        self, pop_visitors: pd.DataFrame, maybe_enroll: pd.Index
+    ) -> pd.DataFrame:
+        """Enrolls simulants in polypill intervention program. It updates
+        the polypill column as well as sbp medication adherence column (the
+        effect of polypill intervention).
+
+        Reminder: 'cat1' polypill means enrolled and 'cat2' means not enrolled.
+        """
+        current = pop_visitors.loc[maybe_enroll, data_values.COLUMNS.POLYPILL]
```
rmudambi: You can replace all of this with:

```
pop_visitors.loc[maybe_enroll, data_values.COLUMNS.POLYPILL] = self.polypill(maybe_enroll)
pop_visitors.loc[maybe_enroll, data_values.COLUMNS.SBP_MEDICATION_ADHERENCE] = self.sbp_medication_adherence(maybe_enroll)
```

This is because of the way polypill acts on the exposure parameters of adherence and not the exposure itself, as we discussed this morning, so it isn't incorrect to call it on people who are already on the intervention.

src/vivarium_nih_us_cvd/model_specifications/nih_us_cvd.yaml
```diff
@@ -69,17 +71,17 @@ configuration:
             month: 1
             day: 1
         observation_start: # TODO: Implement for all observers (currently only custom observers)
-            year: 2023
+            year: 2021
```
rmudambi: I assume these dates are incorrect?

src/vivarium_nih_us_cvd/model_specifications/nih_us_cvd.yaml
```diff
@@ -95,14 +97,12 @@ configuration:
...
-        # being mapped highest to lowest where cat1 is highest risk. Here, an 
-        # outreach.exposure = 0.0 is the highest risk (no outreach means less
-        # likely to take medication) but will map to cat2 (not cat1)
+        exposure: 0
+    polypill:
         exposure: 0
     intervention:
-        scenario: "outreach_50"
-    outreach_scale_up:
+        scenario: "polypill_100"
```
rmudambi: Change the default back to baseline?

rmudambi (changes requested):

rmudambi (approved):

PR #85, opened 2023-02-24
PAF observer

hussain-jafari:
### Description
- *Category*: observer
- *JIRA issue*: [MIC-3822](https://jira.ihme.washington.edu/browse/MIC-3822)

### Changes and notes
I removed observer things that worked with the old results management system because they stopped working when using the new results management VPH branch. I'll be using a heart failure PAF develop branch which works with the new VPH results branch but keep things in main consistent with the main/develop VPH branch.

### Verification and Testing
Put a breakpoint in the aggregator function and checked that relative risk values were around what we would expect for demographic groups (looking at age) and checked that calculated PAF values in output of sim run looked reasonable. Checks were done for BMI and FPG on acute myocardial infarction incidence.


src/vivarium_nih_us_cvd/components/__init__.py
```diff
@@ -2,11 +2,6 @@
 from .effects import InterventionAdherenceEffect
 from .healthcare_utilization import HealthcareUtilization
 from .interventions import LinearScaleUp
-from .observers import (
```
rmudambi, line 5: You shouldn't be deleting all the observers. This ticket is just to create the PAFObserver. If having the other observers present is breaking anything, you can always put the PAFObserver in its own file.

rmudambi (changes requested):

src/vivarium_nih_us_cvd/components/__init__.py
hussain-jafari: The issue is that the observers contain references to the old ResultsStratifier, which contains references to objects which no longer exist in the new results management VPH. If we keep both observers, the code will not be runnable under either version of VPH without some modification.

src/vivarium_nih_us_cvd/components/paf_observer.py
```diff
@@ -0,0 +1,60 @@
...
+        relative_risk = self.risk_effect.target_modifier(x.index, pd.Series(1, index=x.index))
+        mean_rr = relative_risk.mean()
+        paf = (mean_rr - 1) / mean_rr
+
+        return paf
+
+    def get_configuration_defaults(self) -> Dict[str, Dict]:
+        return {
+            "stratification": {
+                self.risk: PAFObserver.configuration_defaults["stratification"]["paf"]
```
rmudambi, line 60: This configuration key should probably be `f"{self.risk}_paf"`

src/vivarium_nih_us_cvd/components/paf_observer.py
```diff
@@ -0,0 +1,60 @@
...
+    def __repr__(self):
+        return f"PAFObserver({self.risk}, {self.target})"
+
+    @property
+    def name(self):
+        return f"paf_observer.{self.risk}.{self.target}"
+
+    # noinspection PyAttributeOutsideInit
+    def setup(self, builder: Builder) -> None:
+        self.step_size = builder.time.step_size()
```
rmudambi: You don't use this.

src/vivarium_nih_us_cvd/components/paf_observer.py
```diff
@@ -0,0 +1,60 @@
...
+        self.config = builder.configuration.stratification["paf"]
```
rmudambi: Since you only use this in the `setup` method, there's no need to save this to the `self` object.

src/vivarium_nih_us_cvd/components/paf_observer.py
```diff
@@ -0,0 +1,60 @@
...
+        self.risk_effect = builder.components.get_component(
+            f"risk_effect.{self.risk}.{self.target}"
+        )
+
+        builder.results.register_observation(
+            name=f"calculated_paf_{self.risk}_on_{self.target}",
+            pop_filter=f'alive == "alive"',
```
rmudambi: This doesn't need to be an f-string

rmudambi (approved):

stevebachmeier (approved):

src/vivarium_nih_us_cvd/components/paf_observer.py
hussain-jafari: I tried running this and realized this refers to the PAFObserver configuration and we don't have access to the risk when those defaults are defined. In the disease and mortality observer examples in VPH, the configuration is stratification.disease and stratification.mortality. I'll update the configuration in the setup method to refer to the risk.

PR #88, opened 2023-03-07
Define PAF model spec

hussain-jafari:
### Description
- *Category*: model spec
- *JIRA issue*: [MIC-3823](https://jira.ihme.washington.edu/browse/MIC-3823)

### Changes and notes
Defining minimal model specification to calculate PAFs. Some other changes which came up:

- subclass RiskEffect to not get PAF data
- define configuration defaults in categorical SBP exposure
- widen start and end bins for converting continuous sbp to categorical
- add categorical sbp distribution to artifact

### Verification and Testing
Ran PAF model and checked that PAFs looked reasonable (less than 1, 0 for expected demographic groups).

### DIFF
--- a/src/vivarium_nih_us_cvd/components/risks.py
+++ b/src/vivarium_nih_us_cvd/components/risks.py
@@ -1,6 +1,7 @@
-import pandas as pd
-import numpy as np
 from typing import Dict
+
+import numpy as np
+import pandas as pd
 from vivarium.framework.engine import Builder
 from vivarium.framework.population.manager import PopulationView
 from vivarium.framework.values import Pipeline
--- a/src/vivarium_nih_us_cvd/components/risks.py
+++ b/src/vivarium_nih_us_cvd/components/risks.py
@@ -1,4 +1,6 @@
 import pandas as pd
+import numpy as np
+from typing import Dict
 from vivarium.framework.engine import Builder
 from vivarium.framework.population.manager import PopulationView
 from vivarium.framework.values import Pipeline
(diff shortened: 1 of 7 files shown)


src/vivarium_nih_us_cvd/model_specifications/heart_failure_pafs.yaml
```diff
@@ -0,0 +1,58 @@
...
+            year: 2021
+            month: 1
+            day: 1
+        end:
+            year: 2021
+            month: 1
+            day: 2
+        step_size: 28 # Days
+    population:
+        population_size: 100_000
```
rmudambi, line 53: Did we get clarification from the RT what population size we want to run with? And if they want to parallelize by age group to guarantee we get enough people in each age group to calculate the PAF?

rmudambi (approved):

src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -148,11 +163,11 @@ def _get_current_exposure(self, index: pd.Index) -> pd.Series:
         continuous_exposure = self.continuous_exposure(index)
 
         bins = [
-            RISK_EXPOSURE_LIMITS["high_systolic_blood_pressure"]["minimum"],
+            0,
```
stevebachmeier, line 166: Would we ever have <0 at this point?

hussain-jafari: Nope, unless something has gone very wrong. But then attempting to call the riskeffect target modifier will throw an error.

src/vivarium_nih_us_cvd/model_specifications/heart_failure_pafs.yaml
hussain-jafari: For our first run Syl said we want 100,000 population size with 25 draws. I brought up parallelizing by age group and it's something she'll consider but for this run we can run without it. I'll ask again.

PR #94, opened 2023-03-27
update ldl ladder to include ascvd score

hussain-jafari:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3685](https://jira.ihme.washington.edu/browse/MIC-3685)
- *Research reference*: [LDL-C Treatment Ramp](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_us_cvd/concept_model.html#healthcare-system-modeling)

### Changes and notes
Add a check for high ASCVD when ramping up medication for currently treated simulants.  

### Verification and Testing
Checked ASVCD scores of simulants whose medications were changing and confirmed that they were not lower than 7.5. Also checked other values were correct for people in that specific ramp.


stevebachmeier (approved):
Told you it would be easy once you found where to actually apply it!

PR #97, opened 2023-03-30
initialize lifestyle columns

hussain-jafari:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3897](https://jira.ihme.washington.edu/browse/MIC-3897)
- *Research reference*: [Last FPG test initialization](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_us_cvd/concept_model.html#initialization-parameters)

### Verification and Testing
In an interactive context at initialization, checked that lifestyle column was NaT and that last FPG test had the right bounds, dates for 71% of eligible simulants, and that these dates were roughly uniformly distributed.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/treatment.py
+++ b/src/vivarium_nih_us_cvd/components/treatment.py
@@ -62,7 +62,7 @@ class Treatment:
             data_values.COLUMNS.OUTREACH,
             data_values.COLUMNS.POLYPILL,
             data_values.COLUMNS.LIFESTYLE,
-            data_values.COLUMNS.LAST_FPG_TEST_DATE
+            data_values.COLUMNS.LAST_FPG_TEST_DATE,
         ]
         columns_required_on_initialization = [
             "age",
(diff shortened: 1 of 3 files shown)


src/vivarium_nih_us_cvd/constants/data_values.py
```diff
@@ -489,7 +493,7 @@ class __OutreachEffectSBP(NamedTuple):
 
     @property
     def name(self):
-        return "sbp_multiplier"
+        return "outreach_effect_sbp"
```
hussain-jafari, line 496: Fixing typo that I noticed.

src/vivarium_nih_us_cvd/model_specifications/nih_us_cvd.yaml
```diff
@@ -13,6 +13,7 @@ components:
             - Risk("risk_factor.ldlc_medication_adherence")
             - Risk("risk_factor.outreach")
             - Risk("risk_factor.polypill")
+            - Risk("risk_factor.lifestyle")
```
rmudambi, line 16: Merge issue? Wasn't this in your previous PR?

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -257,12 +263,48 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
...
+            pop[is_eligible_for_testing],
+            [data_values.FPG_TESTING.PROBABILITY_OF_TESTING_GIVEN_ELIGIBLE]
+            * sum(is_eligible_for_testing),
+        )
+
+        # sample from dates uniformly distributed from 0 to 3 years before sim start date
+        starting_event_time = self.clock()
+        fpg_test_start_time = starting_event_time
+        # Subtract step size until we reach the first day less than or equal to 3 years before our sim starting event date
+        while fpg_test_start_time >= starting_event_time - pd.DateOffset(
```
rmudambi: Why are we doing this like this? Why not just draw a float from a uniform distribution, multiply that value by a `pd.TimeDelta` of 3 years, and subtract that from `self.clock()`? We don't need these times that occurred before the simulation started to occur precisely on a timestep cadence.

rmudambi (changes requested):

src/vivarium_nih_us_cvd/model_specifications/nih_us_cvd.yaml
hussain-jafari: Yeah I just looked at main and this change has already been merged.

src/vivarium_nih_us_cvd/components/treatment.py
hussain-jafari: I'll make this change.

rmudambi (approved):

PR #101, opened 2023-04-04
enroll in lifestyle

hussain-jafari:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3901](https://jira.ihme.washington.edu/browse/MIC-3901)
- *Research reference*: [lifestyle intervention ramp](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_us_cvd/concept_model.html#healthcare-system-modeling)

### Verification and Testing
Ran an interactive sim and checked that simulants were having their lifestyle column updated with the current clock time and that their FPG values were within the expected bounds. Also checked that we had (roughly) the expected ratio of enrolled simulants given they were eligible.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/treatment.py
+++ b/src/vivarium_nih_us_cvd/components/treatment.py
@@ -40,7 +40,9 @@ class Treatment:
         )
         self.outreach = builder.value.get_value(data_values.PIPELINES.OUTREACH_EXPOSURE)
         self.polypill = builder.value.get_value(data_values.PIPELINES.POLYPILL_EXPOSURE)
+        self.lifestyle = builder.value.get_value(data_values.PIPELINES.LIFESTYLE_EXPOSURE)
         self.bmi = builder.value.get_value(data_values.PIPELINES.BMI_EXPOSURE)
+        self.fpg = builder.value.get_value(data_values.PIPELINES.FPG_EXPOSURE)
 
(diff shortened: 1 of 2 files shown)


src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -864,24 +868,44 @@ def apply_lifestyle_ramp(self, pop_visitors: pd.DataFrame) -> pd.DataFrame:
...
+        tested_simulants.loc[newly_enrolled, data_values.COLUMNS.LIFESTYLE] = self.clock()
+
         self.population_view.update(
             pop_visitors[
                 [
                     data_values.COLUMNS.LAST_FPG_TEST_DATE,
                 ]
             ]
         )
+        self.population_view.update(
```
rmudambi, line 902: Nit: make the changes to the Lifestyle enrollment column directly on `pop_visitors` so you only need to make one `update` call.

rmudambi (approved):

src/vivarium_nih_us_cvd/components/treatment.py
hussain-jafari: That was the cleaner approach I mentioned I was going to spend 15 minutes on this morning, but some of the booleans would only get defined properly using the tested simulants subset so I just went with this since it's working.

PR #107, opened 2023-04-14
lifestyle adherence

hussain-jafari:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3905](https://jira.ihme.washington.edu/browse/MIC-3905)
- *Research reference*: [lifestyle dropout](https://vivarium-research.readthedocs.io/en/latest/models/intervention_models/crm_mgmt/vivarium_lifestyle.html#affected-outcomes)

### Verification and Testing
Looked at lifestyle adherence column in an interactive context and checked that ratio of adherent simulants was roughly what we wanted.


src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -377,6 +378,12 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
         )
         pop = self.initialize_medication_coverage(pop)
 
+        # Generate lifestyle adherence column
+        lifestyle_propensity = self.randomness.get_draw(pop_data.index)
```
rmudambi: Don't we need an additional key here?

rmudambi (approved):

src/vivarium_nih_us_cvd/components/treatment.py
hussain-jafari: I was able to get uniformly distributed draws and create an adherence column with the expected ratio without it but I'll add one.

PR #117, opened 2023-05-08
implement poc configuration of ischemic stroke

rmudambi:
### Description

- *Category*: proof-of-concept
- *JIRA issue*: [MIC-4056](https://jira.ihme.washington.edu/browse/MIC-4056)
- *Research reference*: 

### Changes and notes
Black removed some blank lines
Added causes.yaml config file to configure ischemic stroke
Created new config files for running only the cause model in parallel
Commented out results processing code to make it work on the stripped down model

### Verification and Testing
Ran parallel sim for three years and validated using a notebook provided by Syl


src/vivarium_nih_us_cvd/components/causes.py
```diff
@@ -0,0 +1,163 @@
...
+
+    ##################
+    # Helper methods #
+    ##################
+
+    def get_disease_models(self) -> List:
+        disease_models = []
+
+        full_config = ConfigTree(layers=["default", "model_spec"])
+        full_config.update(CAUSE_RISK_CONFIG, layer="model_spec")
```
stevebachmeier, line 43: We should maybe re-think these layer names now since both yamls make up the "model spec"

src/vivarium_nih_us_cvd/components/causes.py
```diff
@@ -0,0 +1,163 @@
...
+        self.add_default_config_layer(full_config)
+        config = full_config.causes
+
+        for cause_name, cause_config in config.items():
+            states: Dict[str, BaseDiseaseState] = {
```
stevebachmeier, line 48: Jeez, we add typing outside of functions so rarely that I had to stare at this for a second

albrja: I didnt know we could do this.

src/vivarium_nih_us_cvd/model_specifications/poc_config.yaml
```diff
@@ -0,0 +1,54 @@
...
+    vivarium_nih_us_cvd:
+        components:
+            - Causes()
+#            - Risks()
+
+configuration:
+    input_data:
+        input_draw_number: 0
+        location: 'Alabama'
+        artifact_path: '/home/rmudambi/scratch/alabama.hdf'
```
mattkappel, line 21: Not sure if you intended to leave this path in here.

src/vivarium_nih_us_cvd/components/causes.py
mattkappel: You could reasonably argue that the causes.yaml is not the[ model_spec, as currently defined.](https://vivarium.readthedocs.io/en/latest/concepts/model_specification/index.html)

src/vivarium_nih_us_cvd/components/causes.py
mattkappel: I'm almost certain I can find code one or both of you've written with this lol.

mattkappel (approved):
It might be useful to add reasons why you've commented out blocks (aside from the PR description)*, but I suppose this is all intermediary as a PoC feature.

src/vivarium_nih_us_cvd/model_specifications/poc_config.yaml
rmudambi: oops good catch

PR #121, opened 2023-05-11
add ability to configure a cause model using YAML

rmudambi:
### Description

- *Category*: proof-of-concept
- *JIRA issue*: [MIC-4056](https://jira.ihme.washington.edu/browse/MIC-4056)
- *Research reference*: 

### Changes and notes
Add ability to configure a cause model using YAML 

### Verification and Testing
Ran parallel simulation and verified that cause model validated


src/vivarium_nih_us_cvd/components/causes/causes.py
```diff
@@ -0,0 +1,152 @@
...
+    SusceptibleState,
+    TransientDiseaseState,
+)
+from vivarium_public_health.utilities import TargetString
+
+from vivarium_nih_us_cvd.constants.paths import CAUSE_RISK_CONFIG
+
+
+class Causes:
+    # todo could modify the component configuration parser plugin
```
mattkappel, line 21: Any Jira needed for this?

src/vivarium_nih_us_cvd/components/causes/causes.yaml
```diff
@@ -0,0 +1,146 @@
...
+#        transitions:
+#            acute_ischemic_stroke_incidence:
+#                source: "susceptible"
+#                sink: "acute_ischemic_stroke"
+#                data_type: "rate"
+#                get_data_functions:
+#                    incidence_rate: None                        # default
+#                    transition_rate: None                       # default
+#                    remission_rate: None                        # default
+#                    proportion: None                            # default
```
mattkappel, line 38: Is it still useful to have this commented block?

mattkappel (approved):
Just some cleanup comments.

src/vivarium_nih_us_cvd/components/causes/causes.py
rmudambi: Yes, good reminder. I'll create a ticket

src/vivarium_nih_us_cvd/components/causes/causes.yaml
rmudambi: I think it's useful until we have some documentation of what defaults are.

hussain-jafari (approved):

hussain-jafari:
Our typical runs have results starting at 2023 but I'm also OK with starting at 2021 (which i saw in post processing and in the yaml for custom observers). Was that to keep all observers consistent?

rmudambi:
> Our typical runs have results starting at 2023 but I'm also OK with starting at 2021 (which i saw in post processing and in the yaml for custom observers). Was that to keep all observers consistent?

I didn't change the "real" model spec, so you can decide what to do there.

PR #128, opened 2023-06-08
better handle existing artifacts

stevebachmeier:
### Description
- *Category*: bugfix
- *JIRA issue*: 
- *Research reference*: 

### Changes and notes
The logic was still incorrect; it was still asking to delete all existing
artifacts when I only want it to delete the specific one I'm trying to build.

### Verification and Testing
tested locally to make sure all combinations of location went into the correct
logic.

### DIFF
--- a/src/vivarium_nih_us_cvd/tools/make_artifacts.py
+++ b/src/vivarium_nih_us_cvd/tools/make_artifacts.py
@@ -47,9 +47,9 @@ def check_for_existing(
             if item.is_file() and item.suffix == ".hdf"
         ]
     )
-    locations = set([sanitize_location(loc) for loc in metadata.LOCATIONS])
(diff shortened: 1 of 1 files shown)


albrja (approved):

mattkappel (approved):

src/vivarium_nih_us_cvd/tools/make_artifacts.py
```diff
@@ -49,8 +49,12 @@ def check_for_existing(
     )
     locations = set([sanitize_location(loc) for loc in metadata.LOCATIONS])
```
rmudambi: Nit: `locations` is only ever used inside the `location == "all"` block, so you can define it inside.

rmudambi (approved):

PR #130, opened 2023-06-15
Bugfix/sbachmei/fix locations with space

stevebachmeier:
### Description
- *Category*: bugfix
- *JIRA issue*: na
- *Research reference*: na

### Changes and notes
This fixes locations w/ spaces in their names (e.g. North Carolina, etc). I've also
added another helper python script to run through all artifacts and ensure that
there are 5 PAF keys (it doesn't to a qualitative check but at least ensures
that the expected number of keys got added)

### Verification and Testing
The helper is what verified that we finally have 51 PAF-ful artifacts! :)


paf_utilities/README
```diff
@@ -0,0 +1 @@
+TBD
```
hussain-jafari, line 1: Assuming this is a placeholder for the documentation :)

hussain-jafari (approved):

paf_utilities/README
stevebachmeier: Yep!

PR #132, opened 2023-06-21
refactor custom cvd observers

hussain-jafari:
### Description
- *Category*: refactor
- *JIRA issue*: [MIC-4094](https://jira.ihme.washington.edu/browse/MIC-4094)

### Changes and notes
Refactor custom CVD observers to use new results management system in VPH. Add new lifestyle intervention observer. Remove custom observation start time. Reorder VPH and repo components so VPH observers comes after causes (we have to do this because of a property of the mortality observer).

### Verification and Testing
Ran simulation for one step and checked that we were seeing expected outputs. Added sbp medication adherence stratification to check results stratifier changes. Checked in interactive sim that values were what we expected for each observer.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/observers.py
+++ b/src/vivarium_nih_us_cvd/components/observers.py
@@ -78,7 +78,9 @@ class ContinuousRiskObserver:
     def _get_configuration_defaults(self) -> Dict[str, Dict]:
         return {
             "stratification": {
-                self.risk: ContinuousRiskObserver.configuration_defaults["stratification"]["risk"]
+                self.risk: ContinuousRiskObserver.configuration_defaults["stratification"][
+                    "risk"
+                ]
             }
         }
 
@@ -172,6 +174,7 @@ class HealthcareVisitObserver:
                 excluded_stratifications=self.config.exclude,
                 when="collect_metrics",
             )
+
     def calculate_visit_counts(self, x: pd.DataFrame) -> int:
         return len(x)
 
@@ -263,7 +266,7 @@ class CategoricalColumnObserver:
 
 class LifestyleObserver(CategoricalColumnObserver):
     def __init__(self):
-        self.column = 'lifestyle'
+        self.column = "lifestyle"
         self.configuration_defaults = self._get_configuration_defaults()
 
     def register_observations(self, builder: Builder) -> None:
@@ -287,10 +290,10 @@ class LifestyleObserver(CategoricalColumnObserver):
         )
 
     def _get_categories(self) -> List[str]:
-        return ['cat1', 'cat2']
+        return ["cat1", "cat2"]
 
     def calculate_exposed_lifestyle_person_time(self, x: pd.DataFrame) -> float:
-        return sum(~(x['lifestyle'].isna())) * to_years(self.step_size())
+        return sum(~(x["lifestyle"].isna())) * to_years(self.step_size())
 
     def calculate_unexposed_lifestyle_person_time(self, x: pd.DataFrame) -> float:
-        return sum(x['lifestyle'].isna()) * to_years(self.step_size())
+        return sum(x["lifestyle"].isna()) * to_years(self.step_size())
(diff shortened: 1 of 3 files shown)


src/vivarium_nih_us_cvd/components/observers.py
```diff
@@ -178,46 +156,32 @@ def name(self):
...
-
-    def on_collect_metrics(self, event: Event):
-        if event.time < self.observation_start_time:
-            return
-        pop = self.population_view.get(event.index, query='alive == "alive"')
+        for visit_type in data_values.VISIT_TYPE:
+            builder.results.register_observation(
+                name=f"healthcare_visits_{visit_type}",
+                pop_filter=f'alive == "alive" and visit_type == "{visit_type}"',
+                aggregator=self.calculate_visit_counts,
```
rmudambi: The default aggregator is `len` so you shouldn't need to specify this. Even if you did need to specify it, you wouldn't need to define `self.calculate_visit_counts` - you could just put `len` in line 169.

src/vivarium_nih_us_cvd/components/observers.py
```diff
@@ -292,74 +257,40 @@ def _get_categories(self) -> List[str]:
 
         return mapping[self.column]
 
-    def _get_population_view(self, builder: Builder) -> PopulationView:
-        columns_required = ["alive", self.column]
-        return builder.population.get_view(columns_required)
+    def calculate_categorical_person_time(self, x: pd.DataFrame) -> float:
+        return len(x) * to_years(self.step_size())
```
rmudambi, line 257: This is the same aggregator function used by the `DiseaseObserver` to calculate person time. Is that function exposed in such a way that you can just call that? That would be preferable.

src/vivarium_nih_us_cvd/components/observers.py
```diff
@@ -259,24 +222,26 @@ def name(self):
...
+                name=f"{self.column}_{category}_person_time",
+                pop_filter=f'alive == "alive" and {self.column} == "{category}"',
+                aggregator=self.calculate_categorical_person_time,
+                requires_columns=["alive", self.column],
+                additional_stratifications=self.config.include,
+                excluded_stratifications=self.config.exclude,
+                when="time_step__prepare",
+            )
 
     def _get_categories(self) -> List[str]:
```
rmudambi: This function shouldn't have an underscore since you want to be able to override it in the LifestyleObserver.

src/vivarium_nih_us_cvd/model_specifications/nih_us_cvd.yaml
```diff
@@ -91,25 +90,21 @@ configuration:
...
             day: 1
-        observation_start: # TODO: Implement for all observers (currently only custom observers)
-            year: 2023
-            month: 1
-            day: 1
         end:
-            year: 2040
-            month: 12
-            day: 31
+            year: 2021
```
rmudambi: This shouldn't have been changed.

rmudambi (changes requested):

src/vivarium_nih_us_cvd/components/observers.py
hussain-jafari: I could import the DiseaseObserver and pass its aggregate_state_person_time method to the aggregator parameter. Would that suffice?

rmudambi: It's not a static method, so I don't think that'll work. We should think about refactoring that slightly so that we have access to it.

src/vivarium_nih_us_cvd/components/observers.py
```diff
@@ -166,16 +166,13 @@ def setup(self, builder: Builder) -> None:
             builder.results.register_observation(
                 name=f"healthcare_visits_{visit_type}",
                 pop_filter=f'alive == "alive" and visit_type == "{visit_type}"',
-                aggregator=self.calculate_visit_counts,
+                aggregator=len,
```
rmudambi: See https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/results/interface.py#L140. You don't need to define the aggregator at all.

rmudambi (approved):
Looks good other than my minor point here.

src/vivarium_nih_us_cvd/components/observers.py
hussain-jafari: Oh yeah you even mentioned in your original comment. I'll remove this line and merge.

PR #142, opened 2023-07-21
risk correlation

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: [MIC-3769](https://jira.ihme.washington.edu/browse/MIC-3769)
- *Research reference*: [Correlation](https://vivarium-research.readthedocs.io/en/latest/models/risk_correlation/sbp_ldlc_fpg_bmi/index.html#correlation)

### Changes and notes
Add correlated risk which does not initialize propensities. Add risk correlation component which initializes correlated propensities for correlated risks.

### Verification and Testing
Ran simulation for a few time steps. Checked that propensities were correlated at initialization as expected by comparing correlation data with calculated Spearman's rank correlation coefficient for a few pairs and age groups.


src/vivarium_nih_us_cvd/components/risks.py
```diff
@@ -74,7 +74,30 @@ def post_processor(exposure, _):
...
+
+    ########################
+    # Event-driven methods #
+    ########################
+
+    def on_initialize_simulants(self, pop_data: SimulantData) -> None:
+        pass
+
+
+class AdjustedRisk(CorrelatedRisk):
```
rmudambi, line 100: Instead of having `CorrelatedRisk` inherit from `DropValueRisk`, you can do something like this:
```
class DropValueRisk(Risk):
    ...

class CorrelatedRisk(Risk):
    ...

class AdjustedRisk(DropValueRisk, CorrelatedRisk):
    ...

class TruncatedRisk(DropValueRisk, CorrelatedRisk):
    ...
```

The benefit of this is that the depth of inheritance is lower, but also it would leave open the ability to define a Risk class that is a `CorrelatedRisk` but not a `DropValueRisk`. It's relatively simple to understand what is going to happen in this case, since the two classes don't override the same function.

Otherwise, you might as well just combine `CorrelatedRisk` and `DropValueRisk` into a single class `CorrelatedDropValueRisk`, rather than inheriting.

src/vivarium_nih_us_cvd/components/risk_correlation.py
```diff
@@ -0,0 +1,101 @@
...
+            ["age"] + self.propensity_column_names
+        )
+
+        builder.population.initializes_simulants(
+            self.on_initialize_simulants,
+            creates_columns=self.propensity_column_names,
+            requires_columns=["age"],
+        )
+
+    ##################
```
rmudambi: Nit: I'd put the helper functions section below the event-driven methods section.

src/vivarium_nih_us_cvd/components/risk_correlation.py
```diff
@@ -0,0 +1,101 @@
...
+        for risk in self.risks:
+            correlation_data[f"{risk.name}_AND_{risk.name}"] = 1
+
+        # add columns with risk pairs switched in column name
+        risk_pairs = [col for col in correlation_data.columns if "AND" in col]
+        switched_risk_pairs = [
+            pair.split("_AND_")[1] + "_AND_" + pair.split("_AND_")[0] for pair in risk_pairs
+        ]
+
+        for original_column, new_column in dict(zip(risk_pairs, switched_risk_pairs)).items():
```
rmudambi: This is cleaner than using a dict: 
```
tuple(zip(risk_pairs, switched_risk_pairs))
```

src/vivarium_nih_us_cvd/components/risk_correlation.py
```diff
@@ -0,0 +1,101 @@
...
+            covariance_matrix = [
+                [
+                    correlation.iloc[i][f"{first_risk.name}_AND_{second_risk.name}"]
+                    for second_risk in self.risks
+                ]
+                for first_risk in self.risks
+            ]
+
+            probit_propensity = np.random.multivariate_normal(
+                [0] * len(self.risks), covariance_matrix, size=len(age_specific_pop)
```
rmudambi: What does the 0 here mean?

src/vivarium_nih_us_cvd/constants/paths.py
```diff
@@ -24,6 +24,7 @@ class __Filepaths(NamedTuple):
     SBP_DISTRIBUTION_WEIGHTS: Path = DATA_ROOT / "sbp_weights.csv"
     BMI_DISTRIBUTION_WEIGHTS: Path = DATA_ROOT / "bmi_weights.csv"
     FPG_STANDARD_DEVIATION: Path = DATA_ROOT / "fpg_standard_deviation.csv"
+    RISK_CORRELATION: Path = DATA_ROOT / "correlation.csv"
```
rmudambi: Nit: rename to `risk_correlation.csv`?

rmudambi (approved):
Looks good! The comment regarding inheritance is not a required change - it was mostly in case you were unaware that you could have done this instead.

src/vivarium_nih_us_cvd/components/risk_correlation.py
hussain-jafari: Those are the means of the distribution.

PR #155, opened 2023-09-18
refactor components in effect.py

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: N/A
- *Research reference*: 

### Changes and notes
- refactor components in effect.py

### Verification and Testing
Ran interactive sim.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/effects.py
+++ b/src/vivarium_nih_us_cvd/components/effects.py
@@ -1,27 +1,18 @@
-from typing import Callable
+from typing import Optional
 
 import pandas as pd
+from vivarium import Component
 from vivarium.framework.engine import Builder
 from vivarium.framework.lookup import LookupTable
 from vivarium.framework.time import get_time_stamp
 from vivarium_public_health.risks.effect import RiskEffect
 
 from vivarium_nih_us_cvd.constants import data_values, scenarios
+from vivarium_nih_us_cvd.constants.scenarios import InterventionScenario
 
 
-class InterventionAdherenceEffect:
+class InterventionAdherenceEffect(Component):
     """A component to model the impact of the intervention risks on medication adherence levels"""
-
-    def __repr__(self):
-        return "InterventionEffect"
-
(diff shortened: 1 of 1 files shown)


src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -107,25 +98,13 @@ def _polypill_sbp_adherence_modifier(
...
+        self, builder: Builder
+    ) -> Optional[LookupTable]:
+        return None
 
-    #################
-    # Setup methods #
-    #################
-
-    # noinspection PyAttributeOutsideInit
-    def setup(self, builder: Builder) -> None:
```
stevebachmeier, line 126: These methods are undefined since the rename - how was this even working?

src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -107,25 +98,13 @@ def _polypill_sbp_adherence_modifier(
...
-        return f"PAFCalculationRiskEffect(risk={self.risk}, target={self.target})"
-
-    ##############
-    # Properties #
-    ##############
-
-    @property
-    def name(self) -> str:
-        return f"paf_calculation_risk_effect.{self.risk}.{self.target}"
+    def get_population_attributable_fraction_source(
```
stevebachmeier, line 101: So the purpose of this component is to behave like a standard risk effect accept it skips the PAF stuff?

rmudambi: It needs to calculate risk effects, which are then used by the PAF calculation observer.

src/vivarium_nih_us_cvd/components/effects.py
rmudambi: It inherits from `RiskEffect`, so these methods are defined. Why do you think they aren't?

hussain-jafari (approved):

mattkappel (approved):

PR #157, opened 2023-09-22
Implement mediation for IHD and ischemic stroke

stevebachmeier:
### Description

- *Category*: implementation
- *JIRA issue*: [MIC-3686](https://jira.ihme.washington.edu/browse/MIC-3686)
- *Research reference*: 
    - *Research reference*:  
    - [bmi](https://vivarium-research.readthedocs.io/en/latest/models/risk_effects/bmi/index.htm#relative-risksl)
    - [fpg](https://vivarium-research.readthedocs.io/en/latest/models/risk_effects/fpg/index.html#relative-risks)

### Changes and notes
 
This is one of two PRs (the second being adding heart failure mediation)

This adds a new MediatedRiskEffect component which inherits from
RiskEffect. It uses a dictionary to loop through any mediators
and applies the scalar equation from the docs.

### Verification and Testing

Ran a few time steps.


src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -108,3 +109,108 @@ def register_target_modifier(self, builder: Builder) -> None:
...
+            & (mediation_factors["affected_entity"] == self.target.name)
+        ]
+
+        def adjust_target(index: pd.Index, target: pd.Series) -> pd.Series:
+            unadjusted_rr = self.unadjusted_rr(index)
+            scaling_factor = pd.Series(1, index=index)
+            for mediator in self.mediators:
+                unadjusted_mediator_rr = self.unadjusted_mediator_rr[mediator](index)
+                not_tmrel_idx = index[
+                    (unadjusted_mediator_rr != 1.0) & (unadjusted_rr != 1.0)
```
rmudambi, line 203: Don't we only care if the mediator RR is 1? That's where we get the division by 0. The math still works out fine if the mediator is not TMREL, but the unadjusted risk is TMREL.

That said, it is more efficient to check for both as you are doing, because the adjustment is guaranteed to do nothing if there is an unadjusted RR of 1. I think it'd be helpful to add a comment explaining why we're checking for each of the TMRELs here.

rmudambi (approved):

src/vivarium_nih_us_cvd/components/effects.py
stevebachmeier: We had a bit of a discussion about this (https://ihme.slack.com/archives/C01N6LFMN3W/p1695338124205609) and Syl requested that we don't adjust if either RR is 1.

PR #158, opened 2023-09-26
heart failure mediation (dummy data)

stevebachmeier:
### Description

- *Category*:  implementation
- *JIRA issue*: [MIC-3686](https://jira.ihme.washington.edu/browse/MIC-3686)
- *Research reference*:  https://vivarium-research.readthedocs.io/en/latest/models/risk_effects/bmi/index.html#risk-outcome-pair-10-heart-failure

### Changes and notes
 

This adds mediation for heart failure using dummy data (really it's the
mean mediation deltas per group that I copied across 1000 draws
and slightly noised - Syl will provide draw-level data soon and I'll PR again).

### Verification and Testing

Ran 1 year of the sim


src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -157,6 +163,7 @@ class MediatedRiskEffect(RiskEffect):
     def setup(self, builder):
         super().setup(builder)
         # Register unadjusted RR pipelines by passing target=1s to the super's target_modifier
+        self.is_target_hf = "heart_failure" in self.target.name
```
rmudambi: Nit: Feel free to ignore, but you could alternatively do `self.target.name.startswith("heart_failure")`

src/vivarium_nih_us_cvd/components/effects.py
```diff
@@ -194,23 +207,28 @@ def adjust_target(index: pd.Index, target: pd.Series) -> pd.Series:
...
-                mf = mediation_factors.loc[
-                    mediation_factors["mediator_name"] == mediator, "value"
-                ].values[0]
-                delta_mediator = np.log(
-                    mf * (unadjusted_rr.loc[not_tmrel_idx] - 1) + 1
-                ) / np.log(unadjusted_mediator_rr.loc[not_tmrel_idx])
-                scaling_factor.loc[not_tmrel_idx] *= (
-                    unadjusted_mediator_rr.loc[not_tmrel_idx] ** delta_mediator
-                )
+                if self.is_target_hf:
```
rmudambi: Instead of putting the if/else inside the `adjust_target` function it is probably cleaner to define separate `adjust_target` functions inside the if/else you already have above. The `adjust_target` functions in the two branches have very little in common with each other, and this way you aren't calling this if statement every time the function is called. I also think it's easier to read that way.

Also, in the HF case there's really no need to filter down to the `not_tmrel_idx`, since we have no division and thus no div/0.

src/vivarium_nih_us_cvd/constants/data_keys.py
```diff
@@ -362,6 +362,9 @@ def log_name(self):
 
 class __Mediation(NamedTuple):
     MEDIATION_FACTORS: TargetString = TargetString("risk.cause.mediation_factors")
+    HF_DELTAS: TargetString = TargetString(
+        "high_body_mass_index_in_adults_on_heart_failure.categorical_high_systolic_blood_pressure.mediation_deltas"
```
rmudambi: A `TargetString` technically just needs to be a string of the form `x.y.z`, but conventionally `x` would be something like "risk_factor", "cause", or "sequela" (as you can tell by the fact the parts are named `type.name.measure`).

src/vivarium_nih_us_cvd/components/effects.py
stevebachmeier: Ah, good idea about the refactor.

@rmudambi  Regarding the filtering to RR>1, though... what about for RRbmi? In that case it mathematically works (unlike the other equations), but it becomes a fundamentally different thing then since for stroke and MI we do NOT adjust if the risk RR (not just the mediator RR) is 1.

rmudambi: Sorry, you're right. We need to filter out the case where the unadjusted BMI == 1, but the reason is different in the HF case than what you've described in the comment above.

rmudambi (approved):
The other changes are not required if you don't want to make them, but it might make it easier to read 🤷

PR #161, opened 2023-09-27
split VI and VCT setup requirements

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: [MIC-4573(https://jira.ihme.washington.edu/browse/MIC-4573)
- *Research reference*: na

### Changes and notes
 

The data requirements for artifact building relies on gbd stuff which is forcing
pandas==1.5.3 whereas we really would like pandas>=2. This splits VCT and
vivarium_inputs into two requirements ("data" for VI and "cluster" for vct).

"dev" continues to include all of them so you'll for now need to
manually install pandas=xxx if you want something else.

NOTE: If you want to build artifacts in parallel, you will need both
"data" and "cluster" since it does indeed rely on the cluster.

NOTE: If we like this then I'll go ahead and update the repo template.

### Verification and Testing

installed "cluster" on my machine and confirmed I did not get VI.


patricktnast (approved):

albrja (approved):

setup.py
```diff
@@ -51,8 +48,9 @@
         install_requires=install_requirements,
         extras_require={
             "test": test_requirements,
-            "data": data_requires,
-            "dev": test_requirements + data_requires,
+            "data": data_requirements,
+            "cluster": cluster_requirements,
```
rmudambi, line 51: As developers we probably want an easy way to install test and cluster without installing data (i.e. be able to build an environment that doesn't rely on the environment limitations of the CC tools), since building artifacts is a smaller part of the dev workflow. 

I have a two proposals:

```python
"test": test_requirements,
"data": data_requirements,
"cluster": cluster_requirements + test_requirements,
"dev": test_requirements + data_requirements + cluster_requirements,
```
or:
```python
"test": test_requirements,
"cluster": cluster_requirements,
"data": data_requirements + cluster_requirements,
"dev": test_requirements + cluster_requirements,
```
If you have another idea feel free to share it.

src/vivarium_nih_us_cvd/tools/make_artifacts.py
```diff
@@ -99,10 +99,8 @@ def build_artifacts(
     verbose
         How noisy the logger should be.
     """
-    import vivarium_cluster_tools as vct
-
     output_dir = Path(output_dir)
-    vct.mkdir(output_dir, parents=True, exists_ok=True)
+    mkdir(output_dir, parents=True, exists_ok=True)
```
rmudambi: Why did you change this import?

src/vivarium_nih_us_cvd/utilities.py
```diff
@@ -188,3 +189,27 @@ def get_random_value_from_normal_distribution(
     """Return a random value assuming normal distribution"""
     draw = randomness.get_draw(index, additional_key=additional_key)
     return stats.norm(loc=mean, scale=sd).ppf(draw)
+
+def mkdir(
+    path: Union[str, Path], umask: int = 0o002, exists_ok: bool = False, parents: bool = False
+) -> None:
+    """Utility method to create a directory with specified permissions.
```
rmudambi: Is this just copy/pasted from VCT?

src/vivarium_nih_us_cvd/tools/make_artifacts.py
stevebachmeier: B/c this was importing vct in make artifacts (the "data" setup requirement")

src/vivarium_nih_us_cvd/utilities.py
stevebachmeier: Yes

setup.py
stevebachmeier: I like the second option better. Though I was just doing `pip install -e .[test, cluster]`

src/vivarium_nih_us_cvd/tools/make_artifacts.py
rmudambi: By using option 2, the `data` requirement includes `cluster`, so you maybe don't need to do this anymore?

stevebachmeier: Oops, yeah good call. I do kinda feel like the `mkdir` utility shouldn't be in vct anyway and should just be a part of the model repos but won't pull that string for now

rmudambi (approved):

PR #166, opened 2023-10-11
implement joint PAF modifier

stevebachmeier:
### Description

- *Category*:  implementation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4583
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/risk_correlation/sbp_ldlc_fpg_bmi/index.html#joint-paf-calculations

### Changes and notes
 

joint PAFs are loaded directly into the artifact now so we no longer need to
calculate them the normal way. To implement, we overwrite the
RiskEffect `register_paf_modifier` to pass and do the custom registration
in the risk_correlation module (perhaps not a perfect spot for it but we needed
somewhere that wasn't registering on a risk-target basis and all these changes
are indeed required due to correlation so that seems better than creating a
new component just to handle this)

The default `list_combiner` and `union_post_processor` used when registering
the PAF pipelines (vph.disease.transition::RateTransition) should still work correctly;
it's just that now we only have a single-item list per target.

### Verification and Testing

ran a couple of time steps.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/risk_correlation.py
+++ b/src/vivarium_nih_us_cvd/components/risk_correlation.py
@@ -4,7 +4,9 @@ import numpy as np
 import pandas as pd
 import scipy
 from vivarium import Component
+from vivarium.framework.artifact import EntityKey
 from vivarium.framework.engine import Builder
+from vivarium.framework.lookup import LookupTable
 from vivarium.framework.population.manager import SimulantData
 from vivarium.framework.randomness import get_hash
 from vivarium_public_health.utilities import EntityString
@@ -14,7 +16,10 @@ from vivarium_nih_us_cvd.constants import paths
 
 
 class RiskCorrelation(Component):
-    """Apply correlation to risk factor propensities"""
+    """Apply correlation to risk factor propensities. It also registers the
+    PAF pipeline modifiers since with correlated risk factors we can no longer
+    assume independent risks when calculating PAFs.
+    """
 
     ##############
     # Properties #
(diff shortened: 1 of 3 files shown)


src/vivarium_nih_us_cvd/data/loader.py
```diff
@@ -1019,9 +1019,6 @@ def load_joint_pafs(artifact_path: str, entity: str, location: str) -> pd.DataFr
     affected_entity_col = pd.Series(
         [output_col.split(".")[-2] for output_col in pafs["index"]]
     )
-    affected_entity_col = affected_entity_col.replace(
```
stevebachmeier, line 1022: This was a bug from the previous PR - we want "heart_failure_residual"

src/vivarium_nih_us_cvd/components/risk_correlation.py
```diff
@@ -47,6 +52,38 @@ def setup(self, builder: Builder) -> None:
...
+    #################
+
+    def get_population_attributable_fraction_source(
+        self, builder: Builder
+    ) -> Dict[str, LookupTable]:
+        paf_data = builder.data.load(
+            "risk_factor.joint_mediated_risks.population_attributable_fraction"
+        )
+        pafs = {}
+        for name, group in paf_data.groupby(["affected_entity", "affected_measure"]):
```
hussain-jafari: If you think it's simpler/cleaner, you can also loop through the groupby using 
`for (name_0, name_1), group in paf_data.groupby()`

hussain-jafari (approved):

rmudambi (approved):

src/vivarium_nih_us_cvd/components/risk_correlation.py
stevebachmeier: ooh yeah I like that more

PR #167, opened 2023-10-12
re-work therapeutic inertia

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4605
- *Research reference*:  PR with changes https://github.com/ihmeuw/vivarium_research/pull/1353/files

### Changes and notes
 

There are four main changes here:
1. Changes to the inertia thresholds
2. Newly-medicated simulants with medium-level sbp can get two types of
  medication now.
3. The calculation of inertia propensity is totally different; it now consists of
  a constant simulant-level component plus a dynamic component.
4. When simulants are considered for treatment changes now occurs 
  at every hospital visit.

NOTE: I think I found a bug w/ the previous implementation. Previously, for
the sbp treatment ramp, newly treated simulants in block B only got put on
medication if they were adherent but I think they were just all supposed to be
put on the treatment (like block C). I've asked for confirmation from Syl.

### Verification and Testing

Ran a few time steps.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/treatment.py
+++ b/src/vivarium_nih_us_cvd/components/treatment.py
@@ -2,6 +2,7 @@ from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
+from scipy import stats
 from vivarium import Component
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
@@ -32,8 +33,8 @@ class Treatment(Component):
             data_values.COLUMNS.OUTREACH,
             data_values.COLUMNS.POLYPILL,
             data_values.COLUMNS.LIFESTYLE,
-            data_values.COLUMNS.SBP_THERAPEUTIC_INERTIA_PROPENSITY,
-            data_values.COLUMNS.LDLC_THERAPEUTIC_INERTIA_PROPENSITY,
+            data_values.COLUMNS.SBP_THERAPEUTIC_INERTIA_CONSTANT_COMPONENT,
+            data_values.COLUMNS.LDLC_THERAPEUTIC_INERTIA_CONSTANT_COMPONENT,
         ]
 
     @property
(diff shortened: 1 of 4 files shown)


src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -755,10 +766,8 @@ def apply_sbp_treatment_ramp(
             pop_visitors[data_values.COLUMNS.SBP_MEDICATION]
             != self.sbp_treatment_map[max(self.sbp_treatment_map)]
         ].index
-        medication_change = (
```
stevebachmeier, line 758: Including `to_prescribe_b` here was a mistake I think. https://ihme.slack.com/archives/C01N6LFMN3W/p1697137488429699

src/vivarium_nih_us_cvd/utilities.py
```diff
@@ -59,34 +59,6 @@ def delete_if_exists(*paths: Union[Path, List[Path]], confirm=False):
             p.unlink()
 
 
-def read_data_by_draw(artifact_path: str, key: str, draw: int) -> pd.DataFrame:
```
stevebachmeier, line 62: this isn't used and seems to be mnch-specific

rmudambi (approved):
Looks good.

Minor note: there are several changes to therapeutic inertia that have been made here. It'd be easier to follow and review if they had been implemented separately and had separate PRs.

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -862,55 +874,40 @@ def apply_ldlc_treatment_ramp(
...
         # elevated LDLC, are not currently medicated, have elevated ASCVD, have
         # no history of MI or IS, and who have high LDLC or ASCVD
         to_prescribe_e = newly_prescribed_young.difference(to_prescribe_d).intersection(
             high_ascvd.union(high_ldlc)
         )
         # [Treatment ramp ID F] Simulants who overcome therapeutic inertia, have
-        # elevated LDLC, are not currently medicated, have elevated ASCVD, have
-        # no history of MI or IS, but who do NOT have high LDLC or ASCVD
-        to_prescribe_f = newly_prescribed_young.difference(to_prescribe_d).difference(
+        # elevated LDLC, are not currently medicated, and are either (1) old or
```
patricktnast, line 899: Nit: is it worth explicitly saying "<75" here? If we think the value might change, it probably shouldn't be hard-coded in ln 854

patricktnast (approved):

PR #168, opened 2023-10-12
support reading causes from external config files

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4597
- *Research reference*: 

### Changes and notes
- Allows for causes to be specified in an external config file within a model repository
- Updates simulation's configuration with the contents of that external file
- Updates the model spec to use the new config parser

### Verification and Testing
Ran CVD model with this parser and verified that components are created correctly.


src/vivarium_nih_us_cvd/plugins/causes_parser.py
```diff
@@ -29,18 +31,33 @@ class CausesConfigurationParser(ComponentConfigurationParser):
...
 
         Note that this method modifies the simulation's component configuration
-        by adding default cause model configuration values to the
-        `component_config` layer and marks states that have multiple exiting
-        transitions with the `is_multi_transition` key.
+        by adding the contents of external configuration files to the
+        `model_override` layer and adding default cause model configuration
+        values for all cause models to the `component_config` layer.
+        It also marks states that have multiple exiting transitions with the
+        `is_multi_transition` key in the `model_override` layer.
```
stevebachmeier, line 60: Update parameter docstring below - component_config is no longer a List

src/vivarium_nih_us_cvd/plugins/causes_parser.py
```diff
@@ -54,17 +71,33 @@ def get_components(
             A list of initialized components.
         """
         components = []
-        if isinstance(component_config, ConfigTree) and "causes" in component_config:
+
+        if not component_config:
+            return components
+
+        if "external_configuration" in component_config:
```
stevebachmeier, line 77: How is it guaranteed that the external_configurations will be parsed prior to the causes loop below?

stevebachmeier (approved):

patricktnast (approved):

src/vivarium_nih_us_cvd/plugins/causes_parser.py
rmudambi: This is guaranteed, because we check for it first here. There isn't any parsing that happens prior to this line.

PR #170, opened 2023-10-16
register paf modifier in new component

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: na
- *Research reference*:  na

### Changes and notes
 
When I tested the mediation stuff last week I must have accidentally been
using an old artifact that included the 
`risk_factor.joint_mediated_risks.popoulation_attributable_fraction` key
during the PAF run (where it shouldn't yet exist). The problem is that
during the PAF run we can't load the PAFs yet (obviously) so this was missed.

This PR makes a new component used just for registering PAF modifiers;
it is included in real sim but NOT the PAF sim.

### Verification and Testing

Ran a few time steps on a shiny new artifact.

### DIFF
--- a/src/vivarium_nih_us_cvd/components/risk_correlation.py
+++ b/src/vivarium_nih_us_cvd/components/risk_correlation.py
@@ -16,10 +16,7 @@ from vivarium_nih_us_cvd.constants import paths
 
 
 class RiskCorrelation(Component):
-    """Apply correlation to risk factor propensities. It also registers the
-    PAF pipeline modifiers since with correlated risk factors we can no longer
-    assume independent risks when calculating PAFs.
-    """
+    """Apply correlation to risk factor propensities"""
 
     ##############
     # Properties #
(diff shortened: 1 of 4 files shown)


src/vivarium_nih_us_cvd/components/risk_correlation.py
```diff
@@ -142,3 +107,49 @@ def update_correlation_data(self, correlation_data: pd.DataFrame) -> pd.DataFram
             correlation_data[new_column] = correlation_data[original_column].values
 
         return correlation_data
+
+
+class JointPAF(Component):
```
stevebachmeier, line 112: No changes to the above deleted code from RiskCorrelation - I simply moved the stuff here

src/vivarium_nih_us_cvd/plugins/causes_parser.py
```diff
@@ -1,9 +1,8 @@
 from importlib import import_module
-
-from pkg_resources import resource_filename
 from typing import Any, Callable, Dict, List, Union
 
 import pandas as pd
+from pkg_resources import resource_filename
```
stevebachmeier: Hmm...why is isort putting this here?

hussain-jafari (approved):

src/vivarium_nih_us_cvd/components/risk_correlation.py
rmudambi: I think a cleaner solution would have been to create a component `PAFCalculationRiskCorrelation(RiskCorrelation)` that overrides these two method with `pass` and including that in the paf calculation sim instead of the normal one.

This is fine though.

rmudambi (approved):

patricktnast (approved):

PR #174, opened 2023-10-18
Extract python version matrix

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4626
- *Research reference*:  na

### Changes and notes
 
This uses the python versions list defined in setup.py to automatically define
the build matrix instead of having them decoupled.


update_versions.py
```diff
@@ -0,0 +1,5 @@
+import json
+
+from setup import max_version, min_version, supported_python_versions
+
+print(json.dumps(supported_python_versions))
```
stevebachmeier: This file will be used later as well to update the README.rst instructions with updated supported Python versions.

src/vivarium_nih_us_cvd/components/observers.py
```diff
@@ -158,10 +158,7 @@ def setup(self, builder: Builder) -> None:
         for visit_type in data_values.VISIT_TYPE:
             builder.results.register_observation(
                 name=f"healthcare_visits_{visit_type}",
-                pop_filter=(
```
stevebachmeier, line 161: :::sigh::: black doing it's thing - ignore this file

src/vivarium_nih_us_cvd/components/treatment.py
```diff
@@ -851,7 +851,9 @@ def apply_ldlc_treatment_ramp(
         above_medium_ldlc = measured_ldlc[
             measured_ldlc >= data_values.LDLC_THRESHOLD.MEDIUM
         ].index
-        old_pop = pop_visitors[pop_visitors["age"] >= data_values.LDLC_OLD_AGE_THRESHOLD].index
```
stevebachmeier, line 854: :::sigh::: black doing it's thing - ignore this file

setup.py
```diff
@@ -2,18 +2,22 @@
 import os
 import sys
 
+from packaging.version import parse
 from setuptools import find_packages, setup
 
-min_version, max_version = ((3, 9), "3.9"), ((3, 11), "3.11")
-if not (min_version[0] <= sys.version_info[:2] <= max_version[0]):
-    # Python 3.5 does not support f-strings
+supported_python_versions = ["3.9", "3.10", "3.11"]
```
stevebachmeier: this isn't strictly necessary but seems easier include only the version strings (rather than the string and the parts tuple like before). 

It also requires we specify all supported versions rather than a min and max because it made it easier to  build the matrix from it - but I'm sure I can figure it out if we truly want a range. The tricky part is what to do if we ever suport different major versions, e.g. 3.11, 3.12, and 4.1. (i.e., I wasn't sure if you said min is 3.11 and max is 4.1, how to know how many versions are between that)

.github/workflows/build.yml
```diff
@@ -1,22 +1,42 @@
...
 jobs:
+  get-python-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.11  # Just used for running the versions script
```
rmudambi: This would need to be maintained manually then? Is there a way to automate this as well?

.github/workflows/build.yml
```diff
@@ -1,22 +1,42 @@
...
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ${{ fromJSON(needs.get-python-versions.outputs.matrix) }}
     defaults:
       run:
         shell: bash -l {0}
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python ${{ "{{" }} matrix.python-version {{ "}}" }}
```
stevebachmeier, line 19: This line is broken and was actually causing these builds to always break. I guess CVD never had builds passing.

setup.py
```diff
@@ -2,18 +2,22 @@
...
+):
     py_version = ".".join([str(v) for v in sys.version_info[:3]])
     error = (
         "\n----------------------------------------\n"
-        "Error: This repo requires python {min_version}-{max_version}.\n"
-        "You are running python {py_version}".format(
-            min_version=min_version[1], max_version=max_version[1], py_version=py_version
-        )
+        f"Error: This repo requires python {min_version.base_version}-{max_version.base_version}.\n"
+        f"You are running python {py_version}."
```
rmudambi: Nit: f-strings are only supported starting in 3.5(?). If you use them here and someone is using an older python they won't get the appropriate message telling them what versions are supported.

update_versions.py
rmudambi, line 1: Is a better name for this file `get_supported_python_versions.py`?

rmudambi: As an alternate approach, can we just store the json list of versions in a file somewhere and simply read it in setup.py and the build script? Seems cleaner and we wouldn't need this python script.

.github/workflows/build.yml
stevebachmeier: Yes, though we were already doing this anyway. I will look into somehow automating this. The problem of course is that you need some version of python installed in order to run the python version getter script.

But that's not really doing anything fancy so my hope was that 3.11 would work for a very long time

setup.py
stevebachmeier: Oh yeah, that's not useful. Will put back

update_versions.py
stevebachmeier: Good idea, though this script will primarily be used to update the README.

mattkappel (approved):
This seems like an improvement. When I was doing the work to add the setuptools_scm business, I imagined/expected that longer term we would move this package metadata to something more canonical than setup.py

mattkappel (approved):

setup.py
```diff
@@ -1,18 +1,30 @@
...
 
+from packaging.version import parse
 from setuptools import find_packages, setup
 
-min_version, max_version = ((3, 9), "3.9"), ((3, 11), "3.11")
-if not (min_version[0] <= sys.version_info[:2] <= max_version[0]):
-    # Python 3.5 does not support f-strings
+f = open("python_versions.json")
+supported_python_versions = json.load(f)
+f.close()
```
rmudambi, line 11: I assume we're not using a context manager for maximal backwards compatibility?

rmudambi (approved):

setup.py
stevebachmeier: Yeah, though support for that probably goes way back? idk, was just kinda afraid to go that route

PR #175, opened 2023-10-18
filter to tracked==True in all observers

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4635
- *Research reference*:  na

### Changes and notes
 
None of the observers were filtering to tracked==True. That resulted in a known
problem with the exposure observers b/c without that filter we were calling
the exposure pipelines and including people who aged out of the sim and so
had expsosures of NaN. But then when sum(column) = NaN if there are any
missingness in column.

Although no issues were flagged in the other observers, I went ahead and
added the filter to all of them.

### Verification and Testing

tbd on a separate run

### DIFF
--- a/src/vivarium_nih_us_cvd/components/observers.py
+++ b/src/vivarium_nih_us_cvd/components/observers.py
@@ -110,7 +110,8 @@ class ContinuousRiskObserver(Component):
 
         builder.results.register_observation(
             name=f"total_exposure_time_risk_{self.risk.name}",
-            pop_filter=f'alive == "alive"',
+            # pop_filter='alive=="alive" and tracked==True',
+            pop_filter='alive=="alive"',
             aggregator=self.aggregate_state_person_time,
             requires_columns=["alive"],
             requires_values=[f"{self.risk.name}.exposure"],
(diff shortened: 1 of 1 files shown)


src/vivarium_nih_us_cvd/components/observers.py
```diff
@@ -110,7 +110,8 @@ def setup(self, builder: Builder) -> None:
 
         builder.results.register_observation(
             name=f"total_exposure_time_risk_{self.risk.name}",
-            pop_filter=f'alive == "alive"',
+            # pop_filter='alive=="alive" and tracked==True',
```
rmudambi: Is this supposed to have tracked==True?

rmudambi (approved):
Do the observers in VPH have this filter?

src/vivarium_nih_us_cvd/components/observers.py
```diff
@@ -217,7 +221,11 @@ def register_observations(self, builder: Builder) -> None:
         for category in self.categories:
             builder.results.register_observation(
                 name=f"{self.column}_{category}_person_time",
-                pop_filter=f'alive == "alive" and {self.column} == "{category}"',
+                pop_filter=(
+                    # 'alive=="alive" and tracked==True and '
```
hussain-jafari: Commenting here in case this should also be uncommented

hussain-jafari (approved):

PR #179, opened 2023-10-27
Feature/sbachmei/auto update python versions in readme

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4672
- *Research reference*: 

### Changes and notes
 
adds a new github workflow to run on PR to main.

NOTE: This has a push to main (the readme update) coming from
github itself. This does mean that when it happens we need to
pull to local.

Also, this _might_ be hiding the workflow status next to the PR.


albrja:
I think its fine as a separate workflow. Im also indifferent but it's cool it adds it as part of the PR to main.

.github/workflows/update_readme.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.11
+      - name: Update README
+        run: |
+          pip install packaging
+          python update_readme.py
+      - name: Commit and push changes
+        run: |
+          git config --local user.email "action@github.com"
```
rmudambi, line 23: Does this make the committer the author of the pull request?

.github/workflows/update_readme.yml
```diff
@@ -0,0 +1,36 @@
...
+          git config --local user.name "github-actions"
+          git diff --quiet && git diff --staged --quiet || (
```
rmudambi, line 25: I don't think it's right for this to commit directly to main, since it's triggered by a pull request targeting main. It might make sense if it was triggered by a commit to main, but since it's triggered by a pull request, it should commit to the branch of the pull request imo.

.github/workflows/update_readme.yml
```diff
@@ -0,0 +1,36 @@
...
+            git add README.rst
+            git commit -am "update README with supported Python versions"
```
rmudambi, line 27: As a side-note, we had discussed PRs automatically committing changes to the changelog, but backed off due to complications about making github perform commits itself. If this works, doesn't that mean we should be able to apply it to that?

rmudambi (changes requested):

.github/workflows/update_readme.yml
stevebachmeier: Ok, I'll need to work on this but basically we want to have this happen on any PUSH (b/c any given push could be modifying the python_versions.json)

.github/workflows/update_readme.yml
stevebachmeier: iirc (I squashed the commits so can't know for sure) the commits came in clearly from github

.github/workflows/update_readme.yml
stevebachmeier: As discussed, I don't see why this wouldn't work on modifying the changlog

.github/workflows/update_readme.yml
stevebachmeier: I take it back - the commits still come from the user (somehow)

rmudambi (approved):

PR #188, opened 2023-12-14
remove unnecessary observers; stop saving out hdf

stevebachmeier:
### Description

- *Category*:  misc
- *JIRA issue*: na
- *Research reference*:  remove unnecessary observers; stop saving out hdf

### Changes and notes
 I forgot to remove the  unnecessary observers for the timing runs; doing that now.

I also found that there's a bug and the transition_count.hdf is somehow corrupted
and unable to be opened (and I suspect this has been a bug for a very long time;
I confirmed it exists in the oldest data we have which is back in August). I confirmed
with Syl that she doesn't ever use the hdfs so I just stopped saving them out.

### DIFF
--- a/src/vivarium_nih_us_cvd/results_processing/process_results.py
+++ b/src/vivarium_nih_us_cvd/results_processing/process_results.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Dict, List, NamedTuple, Union
+from typing import Dict, List, NamedTuple, Optional, Union
 
 import pandas as pd
 import yaml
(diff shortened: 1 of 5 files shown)


vivarium_cluster_tools_launcher_6i5xd_sj.sh
```diff
@@ -0,0 +1,7 @@
+
```
mattkappel: Mistaken inclusion?

src/vivarium_nih_us_cvd/results_processing/process_results.py
```diff
@@ -57,17 +57,17 @@ class MeasureData(NamedTuple):
...
-    intervention_person_time: pd.DataFrame
+    # healthcare_visits: pd.DataFrame
+    # sbp_medication_person_time: pd.DataFrame
+    # ldlc_medication_person_time: pd.DataFrame
+    # intervention_person_time: pd.DataFrame
 
     def dump(self, output_dir: Path):
         for key, df in self._asdict().items():
-            df.to_hdf(output_dir / f"{key}.hdf", key=key)
+            # df.to_hdf(output_dir / f"{key}.hdf", key=key)
```
mattkappel, line 70: Is this broken or are hdf files not desired (for now)?

mattkappel (approved):
Looks good except for the VCT temp script in there.

src/vivarium_nih_us_cvd/results_processing/process_results.py
stevebachmeier: Both. transition_count.hdf is and has always been broken (and Hussain has confirmed he sees it as well in mnch). Syl also confirmed she never reads the hdfs.

